PR #90, opened 2018-10-26
fix the quarter parse failure

hyacker:
I think the expect result is the first day of current quarter's last month, it will be cracked when the month is 1, 2.


lephino:
We're in the process of throwing up a new release, we're going to include this pr in that release.  Hold this pr open for a bit so we don't lose it!

PR #92, opened 2018-10-31
Converted a queue in splunkstream to deque

arctan5x:
Changing a default queue to deque


lephino:
I really don't even like that use.  We should probably rewrite this in the future.

PR #99, opened 2018-11-10
Bugfix/fast 11244

jmeixensperger:
This PR includes a few different fixes, mostly related to the http volume endpoint:
- extra safety checks on volume endpoints for recent exceptions
- refactor of GET /volume on controllers to make sure it isn't returning the GET /conf data
- logic to restart running servers in order to load/use new configurations
- fixed large test for volume endpoint
- some general cleaning and logging adjustments

### DIFF
--- a/splunk_eventgen/eventgen_nameko_controller.py
+++ b/splunk_eventgen/eventgen_nameko_controller.py
@@ -30,6 +30,7 @@ class EventgenController(object):
 
     server_status = {}
     server_confs = {}
+    server_volumes = {}
 
     osvars, config = dict(os.environ), {}
     config["AMQP_HOST"] = osvars.get("EVENTGEN_AMQP_HOST", "localhost")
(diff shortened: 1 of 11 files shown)


splunk_eventgen/eventgen_nameko_controller.py
```diff
@@ -481,6 +482,20 @@ def process_server_confs(self):
         self.server_confs = {}
         return dump_value
 
+    def process_server_volumes(self):
+        current_server_vhosts = self.get_current_server_vhosts()
+        if current_server_vhosts:
+            # Try for 15 iterations to get results from current server vhosts
+            for i in range(15):
```
lephino: why we hard coding this?  Seems like this should be a param somewhere.  Approving, but should probably make a ticket or issue to track replacing it, or add a #TODO statement to mark that this should be a param.

splunk_eventgen/lib/generatorplugin.py
```diff
@@ -37,6 +39,66 @@ def __setstate__(self, d):
...
+                      '_time': time_val}]
+                eventcount += 1
+                self._out.bulksend(l)
+                self._sample.timestamp = None
+            except Exception as e:
+                self.logger.exception("Exception {} happened.".format(type(e)))
+                raise e
+
+        try:
+            endTime = datetime.datetime.now()
```
lephino: Add a #TODO here about making the outputter no lose all the generated events or to put in a try except break.

splunk_eventgen/eventgen_nameko_controller.py
jmeixensperger: I copied this code from another block, so I'm not sure :) I'll make a ticket for it.

PR #117, opened 2018-12-03
add features code and doc

Yangxulight:
1. Create an "outputcounter" to calculate the throughput(event volume and count). This feature can be turned on by setting "outputCounter=true"(default is false) on "eventgen.conf", only work on thread mode. After testing on my laptop, turning on this feature will lead to 1.8% performance degradation.
2. The "<egx_controller_host>/status" api will return the total throughput  of all the egx servers and detail throughput for each server.
3. Update the docs for this feature.


docs/REFERENCE.md
```diff
@@ -230,7 +231,11 @@ maxQueueLength = <maximum items before flushing the queue>
     * Number of items before flushing the output queue
     * Default is per outputMode specific    
 
-
+outputCounter = true | fasle
```
jmeixensperger: typo "fasle"

jmeixensperger:
Looks good, maybe we could add a sample run with this setting into PERFORMANCE.md? We don't really have performance metrics right now.

lephino:
Don't merge this until the typo is fixed.

PR #121, opened 2018-12-24
fix FAST-12048, ts is not correct in sample non-backfill mode

YifengMao:
earliest and latest ts in sample (not backfill mode) doesn't work
1 if earliest/latest is an absolute time, the events always have the timestamp that equals earliest
2 if earlist = -100d and latest = now, the events have the timestamp between -200d and -100d

### DIFF
--- a/splunk_eventgen/lib/eventgentimestamp.py
+++ b/splunk_eventgen/lib/eventgentimestamp.py
@@ -5,7 +5,19 @@ import random
 class EventgenTimestamp(object):
 
     @staticmethod
-    def get_random_timestamp(earliest, latest, sample_earliest, sample_latest):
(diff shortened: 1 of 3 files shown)


arctan5x:
Hi. Looks good overall. Were you able to test out other scenarios? I would recommend testing that with bunch of sample scenarios provided to validate. Let me know how that looks!

YifengMao:
> Hi. Looks good overall. Were you able to test out other scenarios? I would recommend testing that with bunch of sample scenarios provided to validate. Let me know how that looks!

thanks tony. Just tried most samples in tests. It works well, but replay sometimes doesn't work.
it is a simple code issue, fixed in this PR

Issue #126, opened 2019-02-05
token replacement doesn't accept negative floats


technimad:
When using token replacement of type random, the float replacement doesn't accept negative values.

This config is at the moment not valid:
```
token.8.token = meter_8=([0-9.]*)
token.8.replacementType = random
token.8.replacement = float[-3.0:2.0]
```

I see no technical reason to not allow this, so I'll prepare a PR.

PR #127, opened 2019-02-05
Replace token negative float 

technimad:
fixes issue #126 

Negative floats are now allowed, as well as floats without a decimal in the replacement specification.


technimad:
Is the circle ci integration broken? A local `make test` runs ok with no errors reported.

lephino:
checking, sorry for the delay, I missed the email on this update.

lephino:
Circleci wasn't set to build forked builds.  I enabled it for the future.  I also added some reviewers so we can get this looked at.  The only way to retrigger the cicleci build is to close the pr and re-open it.  Don't worry about doing that, we'll forgo that and all future pr's should work.

PR #131, opened 2019-02-26
Fix perdayvolume calculation w/ tokens

jmeixensperger:
perDayVolume doesn't replace any tokens before calculating the size of events. This can lead eventgen to generate the incorrect # of events and can drastically reduce the overall output of events (I saw ~25% less events than expected in one situation).

I also added in the conf files and samples that are being used for the performance metrics.


splunk_eventgen/lib/generatorplugin.py
```diff
@@ -206,16 +165,63 @@ def setupBackfill(self):
...
             self._out.setOutputCounter(output_counter)
         self.gen(count=self.count, earliest=self.start_time, latest=self.end_time, samplename=self._sample.name)
-        #TODO: Make this some how handle an output queue and support intervals and a master queue
+        # TODO: Make this some how handle an output queue and support intervals and a master queue
         # Just double check to see if there's something in queue to flush out at the end of run
         if len(self._out._queue) > 0:
             self.logger.debug("Queue is not empty, flush out at the end of each run")
             self._out.flush()
 
+    def replace_tokens(self, eventsDict, earliest, latest, ignore_tokens=False):
```
lephino: Do we even need ignore_tokens?  What's the usecase for ignoring it?

jmeixensperger: The generatorplugin calls replace_tokens every time it builds events, however the perDayVolume generator now calls replace_tokens before building events. It's preventing us from parsing the sample for tokens more than once, but the code should work fine without it because there will be no tokens left to replace anyways. I would prefer not to use it, if you think the extra parsing won't be costly.

Issue #77, opened 2018-06-07
Cannot generate events in splunk with EventGen


luckyteos:
Hi there,

I just started trying out EventGen, and am following the tutorial from youtube but can't get any events to generate in splunk. The error I get in splunk is "Unable to initialize modular input "modinput_eventgen" defined inside the app "SA-Eventgen": Introspecting scheme=modinput_eventgen: script running failed (exited with code 1).". I have looked through the issues section and online but I still can't find a solution to this issue. Any help would be much appreciated. Thanks :)

[eventgen-stuff.zip](https://github.com/splunk/eventgen/files/2079421/eventgen-stuff.zip)

lephino:
Can you tell me what version you're using?

lephino:
I also see you're using "replay" but you don't have a "replaytimestamp" set, make sure you have one that specifies the time field and where it's at.

luckyteos:
Also I tried out your suggestion, but I think maybe I am not getting the date format correct. Maybe you could advise me as to how you would change the config file to make eventgen generate events? The other error I am getting <error><message>Splunkd daemon is not responding: ('Error connecting to /services/shcluster/config/config: [Errno 10054] An existing connection was forcibly closed by the remote host',)</message></error>. Thanks in advance :)

[eventgen-debug.zip](https://github.com/splunk/eventgen/files/2082656/eventgen-debug.zip)

luckyteos:
Hi there, 

I managed to get EventGen to generate events with the sample tutorial eventgen.conf file provided, though using an earlier version of eventgen -Version 4.0.0. Though I still like to know how I could do it with the latest version 6.0+. Many thanks for any help in advance :)

lephino:
Your regex won't match your timestamp.  I'll shoot you back a sample in a bit.

lephino:
[luckyteos_eventgen_samples.zip](https://github.com/splunk/eventgen/files/2086236/luckyteos_eventgen_samples.zip)
Let me know if you have any questions

lephino:
btw, just install that as a splunk app, and set it to "global" on permissions, then you'll see your time replayed.

luckyteos:
Hi lephino,

Thanks for your help, the app that you gave me is indeed generating events. Though I still have a question, the events generated by replaytimestamp are not that of the current date and time, but instead of the time in the sample file right?

lephino:
[...]

PR #137, opened 2019-03-05
Fix jinja template bug under SA-Eventgen app

li-wu:
1. Fix error "AttributeError: 'NoneType' object has no attribute 'rfind'" when using jinja template under SA-Eventgen app.
2. Fix conf checking error  when starting Splunk for the jinja template settings.
3. Add document to address the permission issue: https://github.com/splunk/eventgen/issues/77
4. Fix a few typos.

### DIFF
--- a/splunk_eventgen/lib/plugins/generator/jinja.py
+++ b/splunk_eventgen/lib/plugins/generator/jinja.py
@@ -1,15 +1,17 @@
 from __future__ import division
-from generatorplugin import GeneratorPlugin
-import datetime, time, os
+import datetime
+import time
+import os
+import random
 try:
     import ujson as json
 except:
     import json as json
-
 from jinja2 import nodes
 from jinja2.ext import Extension
 
-import random
+from generatorplugin import GeneratorPlugin
+
 
 class CantFindTemplate(Exception):
 
@@ -21,6 +23,7 @@ class CantFindTemplate(Exception):
         self.msg = msg
         super(CantFindTemplate, self).__init__(msg)
 
+
 class CantProcessTemplate(Exception):
 
     def __init__(self, msg):
@@ -31,6 +34,7 @@ class CantProcessTemplate(Exception):
         self.msg = msg
         super(CantProcessTemplate, self).__init__(msg)
 
+
 class JinjaTime(Extension):
     tags = set(['time_now', 'time_slice', 'time_delta', 'time_backfill'])
 
@@ -38,7 +42,7 @@ class JinjaTime(Extension):
     def _get_time_slice(earliest, latest, slices, target_slice, slice_type="lower"):
         """
         This method will take a time block bounded by "earliest and latest", and a slice.  It'll then divide the time
-        in sections and return a tuple with 3 arugments, the lower bound, the higher bound, and the target in the middle.
+        in sections and return a tuple with 3 arguments, the lower bound, the higher bound, and the target in the middle.
         :param earliest (in epoch):
         :param latest (in epoch):
         :param slices:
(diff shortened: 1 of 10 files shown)


splunk_eventgen/lib/plugins/generator/jinja.py
```diff
@@ -1,6 +1,8 @@
 from __future__ import division
 from generatorplugin import GeneratorPlugin
-import datetime, time, os
+import datetime
```
GordonWang: To keep a good coding style, A better practice is to load system modules before user project modules.
And put a separation line between system modules and project modules.

splunk_eventgen/lib/plugins/generator/jinja.py
```diff
@@ -1,6 +1,8 @@
...
+import time
+import os
 try:
     import ujson as json
```
GordonWang: this might not be related to this PR. Just wondering if we have no fall back handler for json module loading, what will happen if the json is referred in the below code?

tests/medium/plugins/test_jinja_generator.py
```diff
@@ -0,0 +1,36 @@
...
+
+
+class TestJinjaGenerator(object):
+
+    def test_jinja_generator_to_file(self):
+        configfile = "tests/sample_eventgen_conf/jinja/eventgen.conf.jinja_basic"
+        testargs = ["eventgen", "generate", configfile]
+        file_output_path = os.path.abspath(os.path.join(FILE_DIR, '..', '..', '..', OUTPUT_FILE))
+        # clear the result file
+        open(file_output_path, 'w').close()
```
GordonWang: if it is a file, os.remove will be more readable :)

splunk_eventgen/lib/plugins/generator/jinja.py
lephino: There is a fallback to the generic json library, and if used, you lose the ability to output multiline events.

splunk_eventgen/lib/plugins/generator/jinja.py
```diff
@@ -186,20 +191,34 @@ def gen(self, count, earliest, latest, samplename=None):
...
+            if self.config.configfile:
+                working_dir, working_config_file = os.path.split(self.config.configfile)
+            else:
+                splunk_home = os.environ["SPLUNK_HOME"]
+                app_name = getattr(self._sample, 'app', 'SA-Eventgen')
+                working_dir = os.path.join(splunk_home, 'etc', 'apps', app_name, 'default')
+
             if not hasattr(self._sample, "jinja_template_dir"):
-                template_dir = "templates"
+                template_dir = os.path.join(os.path.dirname(working_dir), 'samples')
```
lephino: I specifically did not want to put this in the samples dir, samples should contain the raw events used for parsing, where this is a new type of file.  I'm alright if we use "samples/templates", but please don't put it in the base of samples.

splunk_eventgen/lib/plugins/generator/jinja.py
```diff
@@ -186,20 +191,34 @@ def gen(self, count, earliest, latest, samplename=None):
                 if self._sample.jinja_count_type in ["line_count", "cycles", "perDayVolume"]:
                     self.jinja_count_type = self._sample.jinja_count_type
             startTime = datetime.datetime.now()
-            working_dir, working_config_file = os.path.split(self.config.configfile)
+
+            # if eventgen is running as Splunk app the configfile is None
+            if self.config.configfile:
+                working_dir, working_config_file = os.path.split(self.config.configfile)
+            else:
+                splunk_home = os.environ["SPLUNK_HOME"]
```
arctan5x: Can we provide a default SPLUNK_HOME in case where environment var "SPLUNK_HOME" doesn't exist?

li-wu: When running as modular input we can always get SPLUNK_HOME. Providing a default value may not apply to both Linux and Windows.

tests/medium/plugins/test_jinja_generator.py
li-wu: changed.

splunk_eventgen/lib/plugins/generator/jinja.py
li-wu: Make sense. Update the default jinja template folder to `<bundle>/samples/templates`.

PR #141, opened 2019-03-06
Feature timeMultiple

jmeixensperger:
Created timer.interval object that is adjusted by timeMultiple. The option can also work with sample mode now.


splunk_eventgen/lib/eventgentimer.py
```diff
@@ -48,6 +47,8 @@ def __init__(self, time, sample=None, config=None, genqueue=None, outputqueue=No
             self.rater = rater_class(self.sample)
             self.generatorPlugin = self.config.getPlugin('generator.' + self.sample.generator, self.sample)
             self.outputPlugin = self.config.getPlugin('output.' + self.sample.outputMode, self.sample)
+            if self.sample.timeMultiple != 1:
```
arctan5x: Let's add a check to see if time multiple is a negative number or 0.

Issue #135, opened 2019-03-04
No generation when count=0


jmeixensperger:
count=0 should cause all events in a sample to be dumped per interval. Instead, it interprets the config the same as count=-1, and no events are generated.
Logic needs to be fixed in eventgentimer.py

PR #142, opened 2019-03-08
[bugfix] when count is 0, should generate all the sample events in file

GordonWang:
This PR fix the issue when running citrix sample data. 
The root cause is that when count is 0, timer will not create generator.
While, the default behavior of eventgen5 is when count is 0, in each interval, the generator should generate number of events according respect to the sample file line count.

But, I am thinking if this issue is not so urgent, I would like to suggest to hold this PR for a while. 
After we creating unit test cases for eventgen, we can be more confident to merge it. 

Now, I am not sure if this fix will introduce some regression bugs. 

@arctan5x @li-wu  What's your thought?


lephino:
the logic in the old code, pre this pr looks more correct.  If I use count=0, I should expect the generator plugin to either be set to perdayvolume and have it's own rater, or that the generator plugin knows how to handle the number of events.  If I hit -1, it should play the whole sample.  I think this clears up a lot of the logic, and chances are the only fix needed is in the logger line (192 of eventgen timer).  We should then tell people who were using 0 to dump a full file to change to "-1"

GordonWang:
> the logic in the old code, pre this pr looks more correct. If I use count=0, I should expect the generator plugin to either be set to perdayvolume and have it's own rater, or that the generator plugin knows how to handle the number of events. If I hit -1, it should play the whole sample. I think this clears up a lot of the logic, and chances are the only fix needed is in the logger line (192 of eventgen timer). We should then tell people who were using 0 to dump a full file to change to "-1"

Thanks for the review. 
For eventgen 6, I think we should provide full compatibility with eventgen5, which means for the same eventgen.conf, eventgen6 should have the same behavior as eventgen5.

In this case, when count is not set or set as 0 in eventgen 5,  it means generating the same number of events as sample file provided in each interval. 
As perdayvolume is the new feature provided in eventgen6, we can find another way to support this without breaking the compatibility. 
I noticed that there is a PR about count=0. https://github.com/splunk/eventgen/issues/135
The behavior here should be changed as well, when count is 0, in each interval, the generator should generate the number of events respect to the sample file events count.
But for the perdayvolume feature, we can ask user to set the perdayvolume parameter to enable it explicitly.

In all, for this PR, I think current changes are not enough to address the behavior of "count=0".We need to set the default count value to 0 as well.

lephino:
100% disagree.  The logic that 0 should create "a full dump" of events on count, vs "-1 dump all the events".  I feel this is something that really doesn't provide clarity staying.  Please change the logic in six to be -1 = full dump, and 0 = disable event generation.  I also think then the default for replay should be "-1" on count and on sample mode it should be "0".   I have no idea when someone uses sample, how many events they want, vs replay, you know they want to copy everything.  Please maintain the logic in 6 that exists today, and fix any bugs around it.

lephino:
Just to clarify, I am aware this differs from eg 5, and I think we should.  The logic behind "0 = dump" all is just a concept that doesn't translate well.  I'm also alright if the default count changes to -1.  This is similar to how eg 6 changes "replaytimestamp".  It's a change that provides logical clarity and is easy for the end user to understand.

jmeixensperger:
> I also think then the default for replay should be "-1" on count and on sample mode it should be "0". I have no idea when someone uses sample, how many events they want, vs replay, you know they want to copy everything.
I think it makes sense to keep the default count -1 for sample mode. If we use 0, we'd see more first-time users asking us "Why am I not seeing any events?".

PR #145, opened 2019-03-13
Windbag generator/count + end=0 edge cases

jmeixensperger:
Patching up some edge cases that I discovered. End = 0 was generating a single interval of events, but it shouldn't generate any now. There is still no default for end, which is what causes it to run indefinitely.
There was also an issue with the windbag generator and count = -1. Since the generator does not use a sample directly, there is no telling what count should actually represent. We will now load the 'windbag' sample if we're using the generator and haven't found any sample yet. This allows count = -1 to continue to rely on the sample size. If users want a specific number of windbag events per interval, count will still work perfectly fine.

### DIFF
--- a/docs/REFERENCE.md
+++ b/docs/REFERENCE.md
@@ -323,6 +323,7 @@ count = <integer>
     * Maximum number of events to generate per sample file
     * -1 means replay the entire sample.
     * Defaults to -1.
+    * When used with the windbag generator, the event count is decided by the sample size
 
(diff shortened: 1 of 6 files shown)


lephino:
I think count -1 is up to the generator to handle.  It should just ignore the count if < 0 or something.  I'd just do a logger.WARN - "count is currently set to -1, there is no sample size with this generator, and this has been overriden to the default count of 60."

PR #147, opened 2019-03-20
add python code style lint and format

li-wu:


Makefile
```diff
@@ -101,3 +103,24 @@ docs:
 
 build_spl: clean
 	python -m splunk_eventgen build --destination ./
+
```
li-wu: `make lint` will only lint the newly added and changed parts of python files here.

Makefile
```diff
@@ -101,3 +103,24 @@ docs:
...
+lint:
+ifeq ($(NEWLY_ADDED_PY_FILES), )
+	@echo 'No newly added python files. Skip...'
+else
+	@flake8 $(NEWLY_ADDED_PY_FILES) || true
+endif
+	@git diff -U0 -- '*.py' | flake8 --diff || true
+
```
li-wu: `make format` will only format newly added python files and  sort imports for both newly added and changed python files.

setup.cfg
```diff
@@ -1,7 +1,12 @@
 [flake8]
-exclude = .git,.tox,__pycache__,env,venv
+exclude = .git,.tox,__pycache__,env,venv,build,dist,docs
 max-line-length = 120
 
 [metadata]
 description-file = README.md
 version-from-file: __init__.py
+
+[yapf]
```
li-wu: `yapf` style is based on `pep8` here.

arctan5x:
@li-wu Will you be including linted code changes to this PR?

li-wu:
@arctan5x , not really. This PR will not include the formatting changes for existing Python code. You can see that the command `make lint` and `make format` only apply to newly added or changed Python files. 
My thought is after we have added enough test cases for existing code, we have a separate PR for the code formatting. Another option is when we fix bug or add new features, we lint/format the code along with the changes.  Any other suggestions?

arctan5x:
@li-wu If you want to introduce linting, I would recommend applying a linter to the codebase all together after the tests cases are added.

GordonWang:
> @li-wu If you want to introduce linting, I would recommend applying a linter to the codebase all together after the tests cases are added.

Totally agree!
We should do this in one commit.

PR #150, opened 2019-03-25
Unittest for config and timeparser module

li-wu:


tests/unit/test_timeparser.py
```diff
@@ -0,0 +1,122 @@
...
+def test_time_delta_2_second(delta, expect):
+    ''' Test timeDelta2secs function, convert time delta object to seconds
+    Normal cases:
+    case 1: time delta is 1 day, expect is 86400
+    case 2: time delta is 1 day 3 hour 15 minutes 32 seconds, expect is 98132
+    case 3: time delta is less than 1 day, only 1 hour 10 minutes, expect is 4200
+    case 4: time delta is 1 hour ago, expect is -3600
+
+    Corner cases:
+    case 1: delta object is None -- invalid input, expect is <TBD>
```
GordonWang: @arctan5x @lephino @jmeixensperger 
I put the case description in the doc string. 
For the time parser help functions, there are some corner cases about empty input argument value, such as empty string or None. Please review if these cases are valid or not. 
Probably for the internal used functions, the None input case is not necessary?

Let me know your opinions about them. Thanks.

tests/sample_eventgen_conf/unit/eventgen.conf.config
```diff
@@ -0,0 +1,38 @@
+[sample]
+sampleDir = tests/sample_eventgen_conf/sample
+outputMode = splunkstream
+splunkMethod = https
+splunkHost = localhost
+splunkPort = 8088
+earliest = -3s
+latest = now
+
+count = 0
```
jmeixensperger: To use the full sample, we need to use count = -1.

tests/unit/test_timeparser.py
```diff
@@ -0,0 +1,122 @@
...
+    '''
+    test timeParserTimeMath function, parse the time modifier
+    Normal Case:
+    Case 1: input "+100s" -- the parser should translate it as 100 seconds later.
+    Case 2: input "-20m" -- the parser should handle the month larger than 12 and translate as 20 months ago
+    Case 3: input '3w' -- the parser should translate as 21 days later.
+
+    Corner Cases:
+    Case 1: input "0s" -- the time parser should return now
+    Case 2: input "123" -- unit is the empty string, behavior <TBD>
```
jmeixensperger: "now" is also a valid input, so I wouldn't say that we expect the same behavior everytime there is no unit. Your example input looks good though.

tests/unit/test_timeparser.py
jmeixensperger: I agree with you @GordonWang, some of the internal functions will never use those types of inputs. I'd suggest making these corner cases for inputs that can be changed by the user's configuration. Out of the cases in this PR marked with <TBD>, I think that the only one we really need is this "case 1" directly above.
For the output, I'm not as sure what we should expect for this case. 0 would be okay with me, as long as we are logging an error message for clarity.

tests/sample_eventgen_conf/unit/eventgen.conf.config
li-wu: changed.

tests/sample_eventgen_conf/unit/eventgen.conf.config
```diff
@@ -0,0 +1,38 @@
+[sample]
+sampleDir = tests/sample_eventgen_conf/sample
+outputMode = splunkstream
```
arctan5x: can we start with HECendpoint and modinput outputmode?

li-wu: This config is not actually used to stream data into Splunk. It is used only for the config unit test.

li-wu:
@arctan5x @lephino @jmeixensperger  I just rebase to the latest change on develop branch. Do you have any other comments for the unit test cases in this PR?

PR #151, opened 2019-03-27
Pep8

jmeixensperger:
There are a lot of changed files here, but no functionality should change because of them (most are spacing changes). I fixed some linting warnings/errors based on Lynch's code formation standards on Confluence. This also includes some automatic formatting with the "yapf" and "isort" packages.
There are a few specific warnings which I ignored (related note in setup.cfg) that can be addressed later. If we continue to ignore them, any CI we set up for code formation should run cleanly.


splunk_eventgen/lib/concurrent/futures/__init__.py
```diff
@@ -1,23 +0,0 @@
-# Copyright 2009 Brian Quinlan. All Rights Reserved.
```
li-wu: `concurrent.futures` is a third party lib that should not be formatted.  And best practice is not checked into the source code.

splunk_eventgen/__main__.py
```diff
@@ -6,71 +6,128 @@
...
 
 import requests
 
+import __init__ as splunk_eventgen_init
+import eventgen_core
+
 FILE_LOCATION = os.path.dirname(os.path.abspath(__file__))
 path_prepend = os.path.join(FILE_LOCATION, 'lib')
 sys.path.append(path_prepend)
-import __init__ as splunk_eventgen_init
```
li-wu: I have not tested whether this will cause import error since the path is inserted into sys path before import.

li-wu:
@jmeixensperger I am not sure whether this formatting will break any functionalities since we do not have enough test cases to cover it.

splunk_eventgen/__main__.py
```diff
@@ -6,71 +6,128 @@
...
-    service_subparser.add_argument("--amqp-webport", type=int, default=None, help="Specify AMQP web port. This can also be set using the environment variable EVENTGEN_AMQP_WEBPORT. Default is 15672")
-    service_subparser.add_argument("--amqp-user", type=str, default=None, help="Specify AMQP user. This can also be set using the environment variable EVENTGEN_AMQP_USER. Default is 'guest'")
-    service_subparser.add_argument("--amqp-pass", type=str, default=None, help="Specify AMQP password. This can also be set using the environment variable EVENTGEN_AMQP_PASS. Default is 'guest'")
-    service_subparser.add_argument("--web-server-address", type=str, default=None, help="Specify nameko webserver address. This can also be set using the environment variable EVENTGEN_WEB_SERVER_ADDR. Default is 0.0.0.0:9500")
+    service_subparser = subparsers.add_parser(
+        'service',
+        help="Run Eventgen as a Nameko service. Parameters for starting this service can be defined as either env" +
+             "variables or CLI arguments, where env variables takes precedence. See help for more info.")
+    service_subparser.add_argument(
```
arctan5x: linter messed it up here :( I actually like the original code. this is harder to read.

splunk_eventgen/__main__.py
```diff
@@ -176,22 +237,21 @@ def rectify_config(config):
         new["AMQP_URI"] = config["AMQP_URI"]
     else:
         if all([config["AMQP_HOST"], config["AMQP_PORT"], config["AMQP_USER"], config["AMQP_PASS"]]):
-            new["AMQP_URI"] = "pyamqp://{user}:{pw}@{host}:{port}".format(user=config["AMQP_USER"],
-                                                                          pw=config["AMQP_PASS"],
-                                                                          host=config["AMQP_HOST"],
-                                                                          port=config["AMQP_PORT"])
+            new["AMQP_URI"] = "pyamqp://{user}:{pw}@{host}:{port}".format(
```
arctan5x: also harder to read

splunk_eventgen/__main__.py
jmeixensperger: Good catch, I'll mark these lines so they don't get reformatted.

splunk_eventgen/__main__.py
jmeixensperger: I disagree for this first 'service' parser because the help is much more readable.
However, I don't think it's necessary to break up every param as its own line (happens more often below this parser).

splunk_eventgen/__main__.py
jmeixensperger: I agree but I'm fine with either way. The problem is that we can only set whether each arg has its own line or not and how long that line can be. Past that, we cannot control how yapf formats this call. There are open feature requests for yapf that would solve this, but they are not currently prioritizing it.

jmeixensperger:
I changed some options for yapf's line breaking and added ignore filters for pep8 rules that would break our functionality. @arctan5x it should be more readable now too.

splunk_eventgen/__main__.py
```diff
@@ -6,71 +6,105 @@
...
-    service_subparser.add_argument("--amqp-port", type=int, default=None, help="Specify AMQP port. This can also be set using the environment variable EVENTGEN_AMQP_PORT. Default is 5672")
-    service_subparser.add_argument("--amqp-webport", type=int, default=None, help="Specify AMQP web port. This can also be set using the environment variable EVENTGEN_AMQP_WEBPORT. Default is 15672")
-    service_subparser.add_argument("--amqp-user", type=str, default=None, help="Specify AMQP user. This can also be set using the environment variable EVENTGEN_AMQP_USER. Default is 'guest'")
-    service_subparser.add_argument("--amqp-pass", type=str, default=None, help="Specify AMQP password. This can also be set using the environment variable EVENTGEN_AMQP_PASS. Default is 'guest'")
-    service_subparser.add_argument("--web-server-address", type=str, default=None, help="Specify nameko webserver address. This can also be set using the environment variable EVENTGEN_WEB_SERVER_ADDR. Default is 0.0.0.0:9500")
+    service_subparser = subparsers.add_parser(
```
arctan5x: this subparser is weird :)

splunk_eventgen/eventgen_nameko_controller.py
```diff
@@ -461,14 +457,15 @@ def receive_volume(self, data):
...
-        server_vhost_len = len(self.server_status) if 'time' not in self.server_status else len(self.server_status)-1
+        server_vhost_len = len(self.server_status) if 'time' not in self.server_status else len(self.server_status) - 1
         if current_server_vhosts:
             for i in range(num_retries):
                 if server_vhost_len != len(current_server_vhosts) or server_time < current_time:
                     time.sleep(delay)
                     current_server_vhosts = self.get_current_server_vhosts()
                     server_time = self.server_status['time'] if 'time' in self.server_status else 0
-                    server_vhost_len = len(self.server_status) if 'time' not in self.server_status else len(self.server_status)-1
+                    server_vhost_len = len(
```
arctan5x: weird

splunk_eventgen/eventgen_nameko_server.py
```diff
@@ -96,9 +96,10 @@ def get_status(self):
         res["CONFIGURED"] = self.eventgen_dependency.configured
         res["CONFIG_FILE"] = self.eventgen_dependency.configfile
         res["TOTAL_VOLUME"] = self.total_volume
-        res["QUEUE_STATUS"] = {'SAMPLE_QUEUE': {'UNFINISHED_TASK': 'N/A', 'QUEUE_LENGTH': 'N/A'},
-                               'OUTPUT_QUEUE': {'UNFINISHED_TASK': 'N/A', 'QUEUE_LENGTH': 'N/A'},
-                               'WORKER_QUEUE': {'UNFINISHED_TASK': 'N/A', 'QUEUE_LENGTH': 'N/A'}}
+        res["QUEUE_STATUS"] = {
```
arctan5x: weird

splunk_eventgen/eventgen_nameko_server.py
```diff
@@ -139,14 +138,8 @@ def index(self):
...
 
-        return home_page.format(host,
-                                eventgen_status,
-                                configured,
-                                config_file,
-                                total_volume,
-                                worker_queue_status,
-                                sample_queue_status,
-                                output_queue_status)
+        return home_page.format(host, eventgen_status, configured, config_file, total_volume, worker_queue_status,
```
arctan5x: not weird but I would prefer the previous code. just easier to read.

splunk_eventgen/eventgen_nameko_server.py
```diff
@@ -375,10 +358,9 @@ def create_new_hec_key(hostname):
                     if new_key:
                         key = create_new_hec_key(formatted_hostname)
 
-                    self.discovered_servers.append({"protocol": str(protocol),
-                                                    "address": str(formatted_hostname),
-                                                    "port": str(hec_port),
-                                                    "key": str(key)})
+                    self.discovered_servers.append({
```
arctan5x: i would prefer the previous code

splunk_eventgen/eventgen_nameko_server.py
```diff
@@ -389,10 +371,9 @@ def create_new_hec_key(hostname):
...
```
arctan5x: i would prefer the previous code

splunk_eventgen/eventgen_nameko_server.py
```diff
@@ -773,10 +745,9 @@ def get_throughput(self):
                 total_count += output_counter.total_output_count
                 throughput_volume += output_counter.throughput_volume
                 throughput_count += output_counter.throughput_count
-            return {'TOTAL_VOLUME_MB': total_volume / (1024*1024),
-                    'TOTAL_COUNT': total_count,
-                    'THROUGHPUT_VOLUME_KB': throughput_volume / (1024),
-                    'THROUGHPUT_COUNT': throughput_count}
+            return {
+                'TOTAL_VOLUME_MB': total_volume / (1024 * 1024), 'TOTAL_COUNT': total_count, 'THROUGHPUT_VOLUME_KB':
```
arctan5x: i would prefer the previous code

splunk_eventgen/identitygen.py
```diff
@@ -4,115 +4,135 @@
...
-	identityGenerator.generate(300000)
-	identityGenerator.getFile(filename="identities.csv",
-		fields=["username","prefix","username","first_name","last_name","suffix","email","phone","phone2","managedBy","priority","bunit","category","watchlist","startDate","endDate","work_city","work_country","work_lat","work_long"],
-		fieldnames = ["identity","prefix","nick","first","last","suffix","email","phone","phone2","managedBy","priority","bunit","category","watchlist","startDate","endDate","work_city","work_country","work_lat","work_long"])
\ No newline at end of file
+    identityGenerator = identityGenerator()
+    identityGenerator.generate(300000)
+    identityGenerator.getFile(
+        filename="identities.csv", fields=[
+            "username", "prefix", "username", "first_name", "last_name", "suffix", "email", "phone", "phone2",
```
arctan5x: this is just weird

splunk_eventgen/lib/concurrent/futures/__init__.py
```diff
@@ -1,19 +1,11 @@
...
-from concurrent.futures._base import (FIRST_COMPLETED,
-                                      FIRST_EXCEPTION,
-                                      ALL_COMPLETED,
-                                      CancelledError,
-                                      TimeoutError,
-                                      Future,
-                                      Executor,
-                                      wait,
-                                      as_completed)
+from concurrent.futures._base import (ALL_COMPLETED, FIRST_COMPLETED, FIRST_EXCEPTION, CancelledError, Executor, Future,
```
arctan5x: i would prefer the previous code

splunk_eventgen/lib/concurrent/futures/process.py
```diff
@@ -299,25 +297,19 @@ def _start_queue_management_thread(self):
...
             self._queue_management_thread = threading.Thread(
-                    target=_queue_management_worker,
-                    args=(weakref.ref(self, weakref_cb),
-                          self._processes,
-                          self._pending_work_items,
-                          self._work_ids,
-                          self._call_queue,
-                          self._result_queue))
+                target=_queue_management_worker,
+                args=(weakref.ref(self, weakref_cb), self._processes, self._pending_work_items, self._work_ids,
```
arctan5x: i would prefer the previous code

splunk_eventgen/lib/eventgenoutput.py
```diff
@@ -22,11 +23,10 @@ def __init__(self, sample):
         self._setup_logging()
         self.output_counter = None
 
-
     def __str__(self):
         """Only used for debugging, outputs a pretty printed representation of this output"""
         # Eliminate recursive going back to parent
-        temp = dict([ (key, value) for (key, value) in self.__dict__.items() if key != '_c'])
+        # temp = dict([(key, value) for (key, value) in self.__dict__.items() if key != '_c'])
```
arctan5x: would this actually be a functional change?

splunk_eventgen/lib/eventgenoutput.py
```diff
@@ -55,18 +55,18 @@ def setOutputCounter(self, output_counter):
...
         """
         Adds msg to the output buffer, flushes if buffer is more than MAXQUEUELENGTH
         """
-        ts = self._sample.timestamp if self._sample.timestamp != None else self._sample.now()
-        self._queue.append({'_raw': msg, 'index': self._sample.index,
-                        'source': self._sample.source, 'sourcetype': self._sample.sourcetype,
-                        'host': self._sample.host, 'hostRegex': self._sample.hostRegex,
-                        '_time': int(time.mktime(ts.timetuple()))})
+        ts = self._sample.timestamp if self._sample.timestamp is not None else self._sample.now()
+        self._queue.append({
```
arctan5x: weird

splunk_eventgen/lib/eventgentimer.py
```diff
@@ -39,23 +41,24 @@ def __init__(self, time, sample=None, config=None, genqueue=None, outputqueue=No
...
                 self.logger.error("Invalid setting for timeMultiple: {}, value should be positive".format(
-                                  self.sample.timeMultiple))
+                    self.sample.timeMultiple))
             elif self.sample.timeMultiple != 1:
                 self.interval = self.sample.interval * self.sample.timeMultiple
                 self.logger.debug("Adjusting interval {} with timeMultiple {}, new interval: {}".format(
-                                  self.sample.interval, self.sample.timeMultiple, self.interval))
-        self.logger.info("Start '%s' generatorWorkers for sample '%s'" % (self.sample.config.generatorWorkers, self.sample.name))
+                    self.sample.interval, self.sample.timeMultiple, self.interval))
+        self.logger.info(
```
arctan5x: weird

splunk_eventgen/lib/eventgentimer.py
```diff
@@ -130,19 +133,15 @@ def real_run(self):
                             backfillnumber += char
                         elif char != "-":
                             backfillletter += char
-                    backfillearliest = timeParserTimeMath(plusminus=mathsymbol,
-                                                          num=backfillnumber,
-                                                          unit=backfillletter,
+                    backfillearliest = timeParserTimeMath(plusminus=mathsymbol, num=backfillnumber, unit=backfillletter,
```
arctan5x: i would prefer the previous code

splunk_eventgen/lib/eventgenoutput.py
jmeixensperger: No, I made this change manually because we don't currently use temp.

arctan5x:
Would you resolve the conflict?

splunk_eventgen/__main__.py
lephino: it's stuff like this that makes me hate line length rules :(

.gitignore
```diff
@@ -17,7 +17,6 @@ stage/
 local/
 eventgen_wsgi.conf
 *.log
-dist
```
li-wu: I think this should not be deleted, right?

setup.cfg
```diff
@@ -1,12 +1,22 @@
 [flake8]
-exclude = .git,.tox,__pycache__,env,venv,build,dist,docs
+exclude = .git,.tox,__pycache__,env,venv,build,dist,docs,splunk_eventgen/lib/concurrent/futures/__init__.py
```
li-wu: The correct third party lib folders to exclude are: `splunk_eventgen/lib/concurrent/futures` and `splunk_eventgen/lib/requests_futures`. right?

.gitignore
jmeixensperger: It was a duplicate line in the file.

setup.cfg
jmeixensperger: Correct, I'll include 'requests_futures' as well.

li-wu: You may also need to revert the changes to the files under these two folders.

splunk_eventgen/eventgen_nameko_controller.py
li-wu: I think for these kinds of long line code , one option is to break the lines instead of formatting it. I make it more readable. 
```
if 'time' not in self.server_status:
    server_vhost_len = len(self.server_status)
else:
    server_vhost_len = len(self.server_status) - 1
```

PR #152, opened 2019-03-28
Fix make docs bug and summary anchor link error

li-wu:
1. Fix `make docs` not working bug;
2. Fix anchors link not working error in SETUP doc page;


docs/README.md
```diff
@@ -0,0 +1,33 @@
+# The Splunk Event Generator
```
li-wu: This file is copied from root folder except the status section.

arctan5x:
Are we setting an internal gitbook for Eventgen? Currently, we use Gitbook pages to automatically do the docs.

li-wu:
@arctan5x . No, `make docs` is used to test the docs on local machine when you change the doc content.

arctan5x:
I understand. Though we don't host on gitbook anymore so we should remove all gitbook related references. I originally left it when we migrated to Github pages but now I wonder if we are going to use Gitbook anymore. Would like to hear your opinion.

li-wu:
I believe GitHub Page(JekyII) has the same functionality as with GitBook(Translate markdown to html). Another option is hosting GitHub Page locally with JekyII. That would require more pre-requirements. (Install Ruby, Gem etc.)

jmeixensperger:
> I believe GitHub Page(JekyII) has the same functionality as with GitBook(Translate markdown to html).
This is my understanding as well. I think we're fine to continue using Github pages.

li-wu:
@arctan5x @jmeixensperger , change `make docs` to host GitHub Page instead of GitBook to test the doc changes locally. That would be consistent with the online GitHub Page.

lephino:
I honestly don't care which doc system we use.  The install + file settings seemed simpler just using gitbook-serve (and less that people had to install).  Honestly leave this up to you guys to make a call on.  I do ask though, if we change this here, we do the same everywhere for our projects.  It's probably just better if we stop using the "make docs" link, and put the link to the github page in the README.MD.

li-wu:
> I honestly don't care which doc system we use. The install + file settings seemed simpler just using gitbook-serve (and less that people had to install). Honestly leave this up to you guys to make a call on. I do ask though, if we change this here, we do the same everywhere for our projects. It's probably just better if we stop using the "make docs" link, and put the link to the github page in the README.MD.

Agree. In this PR, I removed all the gitbook related settings and using GitHub Page per Tony and Jack's comment.  And the `make docs` is just a command to host GitHub Page in local machine and the official GitHub Page is already in `README.MD`.

PR #154, opened 2019-04-08
add extendIndexes feature

Yangxulight:
This PR is to add an "extendIndexes" setting. We want to use this feature to set more indexes for one kind of event.
Here are things I have done:
1. Add  "extendIndexes" setting for each sample and it will be parsed in eventgensamples.py and then add a "index_list" for sample.
2. Generator will choose a index from index_list randomly if you set value to "extendIndexes".

### DIFF
--- a/docs/REFERENCE.md
+++ b/docs/REFERENCE.md
@@ -210,10 +210,10 @@ index = <index>
     * ONLY VALID WITH outputMode SPLUNKSTREAM
     * Splunk index to write events to. Defaults to main if none specified.
     
-extendIndexes = <index_prefix>:<num>,<index2>,<index3>
+extendIndexes = <index_prefix>:<weight>,<index2>,<index3>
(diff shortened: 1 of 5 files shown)


docs/REFERENCE.md
```diff
@@ -210,6 +210,13 @@ index = <index>
     * ONLY VALID WITH outputMode SPLUNKSTREAM
     * Splunk index to write events to. Defaults to main if none specified.
     
+extendIndexes = <index_prefix>:<num>,<index2>,<index3>
+    * Sample level setting. Use this setting enable eventgen to generate multi indexes for one sample. 
+    * if you set the value with pattern like "<index_prefix>:<num>", it will treat <index_prefix> as a prefix of an actual index, 
+    * and <num> is the count of index you want to extend for this sample.
```
arctan5x: this num is actually like a weight. Can we change the wording as a weight?

Yangxulight: Yes, it's better to describe it using the word "weight".

PR #155, opened 2019-04-09
Add sample functional test for replay mode

li-wu:
This PR contains sample functional tests mainly for replay mode including: 
1. normal settings for replay mode;
2. end/timeMultiple/earliest/sampletype setting test;
3. a simple test case for normal sample mode;

Since some test cases are failed as we discussed in the meeting, I do not integrate it into CI process temporarily. Comments are welcome. Thanks.


li-wu:
Close it since it is already merged.

PR #156, opened 2019-04-10
Update docs

li-wu:
I went through the docs and found a few outdated/typo/wrong places.


li-wu:
Another question: does `script input` still apply to current eventgen? http://splunk.github.io/eventgen/BASICS.html#scripted-input

Issue #153, opened 2019-04-08
[BUG] Regression - Running `python ./splunk_eventgen/__main__.py generate splunk_eventgen/README/eventgen.conf.tutorial1` failed


li-wu:
**Description**
Running `python ./splunk_eventgen/__main__.py generate splunk_eventgen/README/eventgen.conf.tutorial1` for `develop` branch failed.

**To Reproduce**
Directly run `python ./splunk_eventgen/__main__.py generate splunk_eventgen/README/eventgen.conf.tutorial1`. 

**Expected behavior**
Give events output. 

**Actual behavior**
Error throw: 
```
  File "./splunk_eventgen/__main__.py", line 25, in <module>
    EVENTGEN_VERSION = splunk_eventgen_init.__version__
AttributeError: 'module' object has no attribute '__version__'
```

**Sample files and eventgen.conf file**
`splunk_eventgen/README/eventgen.conf.tutorial1`

**Additional context**
This is caused by the the `pep8` pull request here: 
https://github.com/splunk/eventgen/pull/151/files#diff-afc41a2be8638faf5aa07904e026e238R17

When moving `import eventgen_core` to the front which inserts `splunk_eventgen/lib` to the front of `sys.path`. Then the `import __init__ as splunk_eventgen_init` imports `splunk_eventgen/lib/__init__.py` which is not expected.

jmeixensperger:
Tested fix locally with reproduction steps.

PR #157, opened 2019-04-10
Post-PEP8 Fixes

jmeixensperger:
Fix previous code formation changes for #153
Also added yapf ignore file for third party libs and re-ran yapf/isort.


.yapfignore
```diff
@@ -0,0 +1,2 @@
+splunk_eventgen/lib/concurrent
```
GordonWang: Can we use pip to install the 3rd party lib instead of putting all the code into our code repo?

li-wu: I have a story to track this.

jmeixensperger: Ok, go ahead and remove this file once that's done then.

PR #163, opened 2019-04-16
Issue 160

arctan5x:


splunk_eventgen/lib/eventgentimer.py
```diff
@@ -111,6 +111,8 @@ def real_run(self):
         if self.end and int(self.end) == 0:
             self.logger.info("End = 0, no events will be generated for sample '%s'" % self.sample.name)
             end = True
+        elif int(self.end) == -1:
```
li-wu: if `self.end` is `None`, this will crash.

Issue #167, opened 2019-04-17
[BUG] default jinja template dir is not correctly resolved when sampleDir is set in conf file


GordonWang:
**Describe the bug**
As the title. when setting sampleDir in sample stanza, the jinja template dir is not resolved correctly.

**To Reproduce**
1. use this conf stanza
```
[test_jinja_template_customized_dir]
end = 1
count = 1
sampleDir = tests/large/sample
generator = jinja
jinja_template_dir = tests/large/sample/templates_4_test
jinja_target_template = test_jinja_tmpl_dir.j2
jinja_variables = {"large_number":10}
earliest = -3s
latest = now
outputMode = stdout
```
2. create a sub dir in "tests/large/sample" named "templates"
3. put template files into templates dir.
4. run eventgen with the conf, error is thrown
```
2019-04-17 17:57:12 eventgen        ERROR    MainProcess Error loading sample file for sample 'test_jinja_template_simple'
2019-04-17 17:57:12 eventgen        ERROR    MainProcess test_jinja_tmpl_simple.j2
Traceback (most recent call last):
  File "/Users/gwang/.conda/envs/eventgen/lib/python2.7/site-packages/splunk_eventgen/lib/plugins/generator/jinja.py", line 218, in gen
    jinja_loaded_template = jinja_env.get_template(str(self._sample.jinja_target_template))
  File "/Users/gwang/.conda/envs/eventgen/lib/python2.7/site-packages/jinja2/environment.py", line 830, in get_template
    return self._load_template(name, self.make_globals(globals))
  File "/Users/gwang/.conda/envs/eventgen/lib/python2.7/site-packages/jinja2/environment.py", line 804, in _load_template
    template = self.loader.load(self, name, globals)
  File "/Users/gwang/.conda/envs/eventgen/lib/python2.7/site-packages/jinja2/loaders.py", line 113, in load
    source, filename, uptodate = self.get_source(environment, name)
  File "/Users/gwang/.conda/envs/eventgen/lib/python2.7/site-packages/jinja2/loaders.py", line 187, in get_source
    raise TemplateNotFound(template)
TemplateNotFound: test_jinja_tmpl_simple.j2
```

**Expected behavior**
default templates dir should be "tests/large/sample/templates"

**Actual behavior**
template not found

**Screenshots**
N/A

**Sample files and eventgen.conf file**
See describe section

**Do you run eventgen with SA-eventgen?**
No

** using eventgen with pip module mode (please complete the following information):**
[...]

PR #176, opened 2019-04-22
Default template dir

jmeixensperger:
Use sampleDir to define a default templates path and load it into our filesystem.
Addresses #167.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -47,7 +47,7 @@ setup(
         'httplib2',
         'jinja2',
         'pyrabbit==1.1.0',
-        'urllib3==1.23',
+        'urllib3>=1.24.2',
         'pyOpenSSL',
         'flake8>=3.7.7',
         'yapf>=0.26.0',
--- a/splunk_eventgen/lib/plugins/generator/jinja.py
+++ b/splunk_eventgen/lib/plugins/generator/jinja.py
@@ -195,8 +195,12 @@ class JinjaGenerator(GeneratorPlugin):
                 app_name = getattr(self._sample, 'app', 'SA-Eventgen')
                 working_dir = os.path.join(splunk_home, 'etc', 'apps', app_name, 'default')
 
+            default_template_dir = os.path.join(os.path.dirname(working_dir), 'samples', 'templates')
+            if hasattr(self._sample, "sampleDir"):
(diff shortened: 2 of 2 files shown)


splunk_eventgen/lib/plugins/generator/jinja.py
```diff
@@ -195,8 +195,12 @@ def gen(self, count, earliest, latest, samplename=None):
                 app_name = getattr(self._sample, 'app', 'SA-Eventgen')
                 working_dir = os.path.join(splunk_home, 'etc', 'apps', app_name, 'default')
 
+            default_template_dir = os.path.join(os.path.dirname(working_dir), 'samples', 'templates')
+            if hasattr(self._sample, "sampleDir"):
```
GordonWang: I have fixed #176 in my local branch and will submit the PR soon. 

I think this only fixed part of the issue. for #176, we need to fix 2 parts
1. resolve the jinja_template_dir with sampleDir when it is a relative path
2. use the jinja_template_dir directly when it is a absolute path.

GordonWang: This issue is addressed in https://github.com/splunk/eventgen/pull/177

Please review this.

setup.py
```diff
@@ -47,7 +47,7 @@ def readme():
         'httplib2',
         'jinja2',
         'pyrabbit==1.1.0',
-        'urllib3==1.23',
+        'urllib3>=1.24.2',
```
li-wu: Also this should be `urllib3==1.24.2`, or it will break all the functional tests.

splunk_eventgen/lib/plugins/generator/jinja.py
li-wu: @jmeixensperger could we just use #177 since it addressed both the `default template dir` issue and `vulnerability ` issue?

GordonWang:
#177 has been merged.

arctan5x:
Please resolve merge conflicts.

li-wu:
@arctan5x @jmeixensperger I think both the `default template dir` issue and the `vulnerability` issue are already fixed at #177 with test cases. So this PR can be closed instead of merging to dev branch.

PR #186, opened 2019-04-29
Issue 159

arctan5x:
TimeMultiple Fix


docs/REFERENCE.md
```diff
@@ -295,9 +295,8 @@ autotimestamp = <boolean>
     * Will enable autotimestamp feature which detects most common forms of timestamps in your samples with no configuration.
 
 timeMultiple = <float>
-    * Will slow down the replay of events by <float> factor. For example, allows a 10 minute sample
-      to play out over 20 minutes with a timeMultiple of 2, or 60 minutes with a timeMultiple of 6.
-      By the converse, make timeMultiple 0.5 will make the events run twice as fast.
+    * Only valid in mode = replay
```
jmeixensperger: Why did we lose support with sample mode?

lephino: I don't think it's that we lost support, it's just that it didn't make sense for it.  If you want to adjust the volume, you'd just change count / perdaysize, not the multiple.

PR #194, opened 2019-05-06
Update index.md

huntershen:
* Ability to be executed inside of Splunk (relying on a comment event generation framework) as well as outside of Splunk
[Hunter] The term "Splunk" here is very vague. What does it refer to - Splunk Enterprise? Splunk Cloud? Splunk Light? Please be more specific. 
.... by allowing generators to replay real data replacing current time by generating data exactly at the same time =intervals as the original data
[Hunter] This is quite incomprehensible - not sure what it means.

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,6 +1,6 @@
 # What is Eventgen?
 
-Splunk Event Generator (Eventgen) is a utility that helps users easily build real-time event generators and elimintes the need for one-off, hard-coded event generators. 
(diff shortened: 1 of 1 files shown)


docs/index.md
```diff
@@ -1,15 +1,14 @@
 # What is Eventgen?
 
-**The Splunk Event Generator is a utility which allows its users to easily build real-time event generators.**
+Splunk Event Generator (Eventgen) is a utility that helps users easily build real-time event generators and elimintes the need for one-off, hard-coded event generators.
```
li-wu: typo `elimintes` => `eliminates`

PR #177, opened 2019-04-23
add test for jinja template test

GordonWang:
- clean up the logic about sampleDir and jinja_template_dir setting
- Add functional test for jinja template generator
- Fixes #174
- Fixes #167

PR #210, opened 2019-05-15
Add 3rd lib in app

li-wu:


splunk_eventgen/splunk_app/README/eventgen.conf.spec
```diff
@@ -22,7 +22,7 @@
 [global]
 disabled = false
 debug = false
-verbose = false
+verbosity = false
```
li-wu: Make the spec consistent for this change: https://github.com/splunk/eventgen/pull/191

splunk_eventgen/lib/plugins/generator/jinja.py
```diff
@@ -193,8 +193,7 @@ def gen(self, count, earliest, latest, samplename=None):
             if self._sample.splunkEmbedded is True:
                 splunk_home = os.environ["SPLUNK_HOME"]
                 app_name = getattr(self._sample, 'app', 'SA-Eventgen')
-                sample_dir = os.path.join(splunk_home, 'etc', 'apps', app_name,
-                                          'default', self._sample.DEFAULT_SAMPLE_DIR)
+                sample_dir = os.path.join(splunk_home, 'etc', 'apps', app_name, 'samples')
```
li-wu: Fix bug introduced in this pull request: https://github.com/splunk/eventgen/pull/177

splunk_eventgen/__main__.py
```diff
@@ -308,6 +308,13 @@ def build_splunk_app(dest, source=os.getcwd(), remove=True):
     directory_default_dir = os.path.join(directory, 'default', 'eventgen.conf')
     eventgen_conf = os.path.join(module_path, 'default', 'eventgen.conf')
     shutil.copyfile(eventgen_conf, directory_default_dir)
+
```
li-wu: Add `ujson` and `jinja2` dependency when building the app.

lephino: Are we sure we want to agree for the install of these libraries for the end user?

splunk_eventgen/__main__.py
```diff
@@ -308,6 +308,13 @@ def build_splunk_app(dest, source=os.getcwd(), remove=True):
...
+    # install 3rd lib dependencies
+    install_target = os.path.join(directory, 'lib')
+    install_cmd = "pip install jinja2 ujson --upgrade --no-compile --no-binary :all: --target " + install_target
```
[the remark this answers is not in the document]
li-wu: changed to use `requirements.txt`

splunk_eventgen/__main__.py
li-wu: Both are BSD license based and I will check it with legal.

jmeixensperger: Any update here?

li-wu: The process with OSS will start with all the 3rd party libs used by Eventgen.  Not just these two in app.

Issue #209, opened 2019-05-15
[FEATURE/IMPROVEMENT] support httpevent output for splunk metric index


YifengMao:
**Is your feature request related to a problem? Please describe.**
current httpevent outputer cannot send neither raw metric events nor well-formed hec payload for splunk metric index.

**Describe the solution you'd like**
add a index type for outputMode = httpevent, default is normal:
splunkIndexType=metric

**Describe alternatives you've considered**
just provide a way for httpevent outputer that send the data from generatorworker as the whole body of http request without putting it into a dictionary. That is, data sample is responsible for putting source/sourcetype/_time into body.

**Additional context**
a well formed hec body for metric index:
{"time": 1557888742.000000, "index":"v3index","source":"metrics","sourcetype":"splunkd","host":"10.1.48.134","event":"m","fields":{"os":"centos", "arch": "x86","service":"15","service_version":"2","service_environment":"test", "metric_name:thread": 18}}

liyancode:
@Yangxulight has done a fix in my local environment which is verified and works well. 

Example data sent:

`{"index":"metric_idx_hec_eg_test","sourcetype":"eventgen","fields":{"service_environment":"test","metric_name":"thread","service":"6","service_version":"1","os":"ubuntu","_value":"12","arch":"x64"},"source":"eventgen","host":"127.0.0.1","time":1557911946,"event":"metric"}`

It is a good fix for my case.

arctan5x:
@Yangxulight Would you verify this issue and close if done?

djschaap:
What is the current status of metric support?  The commit message for release 6.4.0 indicates these metric-related commits were included, yet #226 reverted #215 because of a "new issue".

I would love to have support for metrics - what is this new issue, and can I help fix it?

li-wu:
commit #215 introduced regression issue so we did not include it in the release 6.4.0.  We will fix it soon.

PR #215, opened 2019-05-21
Metrics output plugin

Yangxulight:
This is for requirement: https://github.com/splunk/eventgen/issues/209
What I have done is:
1. Copy httpevent.py plugins and edit it to mertric_httpevent.py
2. Change the http events structure to make it become splunk metric events.


splunk_eventgen/lib/plugins/output/metric_httpevent.py
```diff
@@ -0,0 +1,295 @@
+from __future__ import division
+
+import logging
+import random
+import urllib
+
+from outputplugin import OutputPlugin
```
lephino: Why don't we bust out the "core" of httpevent into a object, and then set both plugins to inherit from the core?

arctan5x: Agree. A lot of duplicate code.

Issue #197, opened 2019-05-06
[BUG] 6.3.5 eventgen CLI multiprocess mode doesn't work


YifengMao:
**Describe the bug**
baseline 6.3.0
CMD:
`splunk_eventgen generate /usr/local/lib/python2.7/dist-packages/splunk_eventgen/default/eventgen_wsgi.conf --multiprocess >/dev/null 2>&1&`

config:

>  [global]
> 
> disabled = false
> debug = false
> verbose = false
> spoolDir = $SPLUNK_HOME/var/spool/splunk
> spoolFile = <SAMPLE>
> breaker = [^\r\n\s]+
> mode = sample
> sampletype = raw
> interval = 60
> delay = 0
> timeMultiple = 1
> count = -1
> earliest = now
> latest = now
> randomizeEvents = false
> outputMode = httpevent
> fileMaxBytes = 10485760
> fileBackupFiles = 5
> splunkPort = 8089
> splunkMethod = https
> index = main
> source = eventgen
> sourcetype = eventgen
> host = 127.0.0.1
> generator = default
> rater = config
> generatorWorkers = 1
> outputWorkers = 1
> timeField = _raw
> threading = thread
> profiler = false
> maxIntervalsBeforeFlush = 3
> maxQueueLength = 0
> useOutputQueue = false
[...]

PR #218, opened 2019-05-22
Bugfix/197 multiprocess not working

li-wu:
This pr fixes issue #197 .  `multiprocess` only works on version `6.3.4`.  And it causes great performance downgrade.  This is caused by previous three fixes: 
1. https://github.com/splunk/eventgen/pull/151/files :   changes to `httpevent.py`
2. https://github.com/splunk/eventgen/pull/189/files
3. https://github.com/splunk/eventgen/pull/162/files

Detail comments are listed below.


splunk_eventgen/lib/eventgentimer.py
```diff
@@ -32,7 +32,7 @@ def __init__(self, time, sample=None, config=None, genqueue=None, outputqueue=No
         self.profiler = config.profiler
         self.config = config
         self.sample = sample
-        self.end = getattr(self.sample, "end") if getattr(self.sample, "end") is not None else -1
```
li-wu: This is introduced in this pr: https://github.com/splunk/eventgen/pull/189/files.  `getattr` will throw exception when `end` is not defined in `self.sample`.

splunk_eventgen/lib/eventgentimer.py
```diff
@@ -187,12 +187,11 @@ def real_run(self):
                             for worker_id in range(self.config.generatorWorkers):
                                 # self.generatorPlugin is only an instance, now we need a real plugin. Make a copy of
                                 # of the sample in case another generator corrupts it.
-                                copy_sample = copy.copy(self.sample)
```
li-wu: I am not sure why we need to make a copy of the example here since for each sample we put a generator object into queue and a  work will pick it up and process it.  Previous fix (https://github.com/splunk/eventgen/pull/162/files) has great performance downgrade for multiprocess mode.

splunk_eventgen/lib/plugins/output/httpevent.py
```diff
@@ -171,15 +171,15 @@ def _sendHTTPEvents(self, payload):
         numberevents = len(payload)
         self.logger.debug("Sending %s events to splunk" % numberevents)
         for line in payload:
-            self.logger.debug("line: %s " % line)
```
li-wu: I am not sure why changing it back to `debugv` here works for historical reason? But changing this back to `debugv` has the right data volume generated for `multiprocess` mode.

splunk_eventgen/lib/eventgentimer.py
lephino: Think this is more from legacy.  I'm guessing this comes from the old eg 5 days.  Think I had to do this when sample's weren't fully loaded in mem, or you'd have 1 sample save a file into a new sample.

Issue #219, opened 2019-05-23
[BUG]6.3.6 eventgen can't start on service mode


Yangxulight:
**Describe the bug**
When I call start rest api from egx controller, the egx server can't start because of missing 'workerQueue'. 

**To Reproduce**
1. create a service mode egx deployment.
2. call setup api and donwload bundle api
3. call start api
4. See error log in 'eventgen-server.log' in egx servers. 

**Expected behavior**
Egx servers run and generate data.

**Actual behavior**
A clear and concise description of what happens after doing the reproduce steps.

**Screenshots**
```
{"timestamp": "2019-05-23 02:27:51", "message": "start method called. Config is /usr/lib/python2.7/site-packages/splunk_eventgen/default/eventgen_wsgi.conf", "level": "INFO"}
{"timestamp": "2019-05-23 02:27:51", "message": "'EventGenerator' object has no attribute 'workerQueue'", "level": "ERROR"}
{"timestamp": "2019-05-23 02:28:15", "message": "Status method called.", "level": "INFO"}
{"CONFIG_FILE": "/usr/lib/python2.7/site-packages/splunk_eventgen/default/eventgen_wsgi.conf", "level": "INFO", "timestamp": "2019-05-23 02:28:15", "CONFIGURED": true, "THROUGHPUT_STATUS": {"TOTAL_VOLUME_MB": 0, "TOTAL_COUNT": 0, "THROUGHPUT_COUNT": 0, "THROUGHPUT_VOLUME_KB": 0}, "EVENTGEN_STATUS": 0, "EVENTGEN_HOST": "egx1", "QUEUE_STATUS": {"WORKER_QUEUE": {"QUEUE_LENGTH": "N/A", "UNFINISHED_TASK": "N/A"}, "SAMPLE_QUEUE": {"QUEUE_LENGTH": "N/A", "UNFINISHED_TASK": "N/A"}, "OUTPUT_QUEUE": {"QUEUE_LENGTH": "N/A", "UNFINISHED_TASK": "N/A"}}, "TOTAL_VOLUME": 0.0}
```

**Sample files and eventgen.conf file**
http://10.66.136.173/web/egx/download/bundle/srv_perfharness/longevity_syslog.tgz

**Do you run eventgen with SA-eventgen?**
No

**If you are using SA-Eventgen with Splunk (please complete the following information):**
 - OS: [e.g. Windows/Mac OS 10.14]
 - Eventgen Version [e.g. 22]
[...]

PR #220, opened 2019-05-24
fix issue 219

li-wu:


li-wu:
Issue #219 is a regression bug introduced in this pull request: https://github.com/splunk/eventgen/pull/207.

PR #221, opened 2019-05-25
Issue 201

arctan5x:
users can specify httpeventAllowFailureCount for httpevent.

If certain url endpoint fails for httpeventAllowFailureCount times, we will remove it from the serverPool


splunk_eventgen/lib/plugins/output/httpevent.py
```diff
@@ -98,28 +98,37 @@ def updateConfig(self, config):
...
             if hasattr(self.config, 'httpeventMaxPayloadSize') and self.config.httpeventMaxPayloadSize:
                 self.httpeventmaxsize = self.config.httpeventMaxPayloadSize
             else:
                 if hasattr(self._sample, 'httpeventMaxPayloadSize') and self._sample.httpeventMaxPayloadSize:
                     self.httpeventmaxsize = self._sample.httpeventMaxPayloadSize
                 else:
                     self.httpeventmaxsize = 10000
+
+            if hasattr(self.config, 'httpeventAllowFailureCount') and self.config.httpeventAllowFailureCount:
+                self.httpeventAllowFailureCount = int(self.config.httpeventOutputMode)
```
li-wu: this should not be `httpeventOutputMode`.

splunk_eventgen/lib/plugins/output/httpevent.py
```diff
@@ -182,50 +181,82 @@ def _sendHTTPEvents(self, payload):
...
-                totalbytessent += len(stringpayload)
-                self.logger.debug(
-                    "End of for loop hit for sending events to splunk, total bytes sent: %s ---- out of %s -----" %
-                    (totalbytessent, totalbytesexpected))
                 self._transmitEvents(stringpayload)
-            except Exception as e:
-                self.logger.exception(str(e))
-                raise e
+                totalbytessent += len(stringpayload)
+                currentreadsize = 0
```
li-wu: the `currentreadsize` should equal to `targetlinesize`? since `targetline` is not sent here.

splunk_eventgen/lib/plugins/output/httpevent.py
```diff
@@ -170,6 +168,7 @@ def _sendHTTPEvents(self, payload):
         totalbytessent = 0
         numberevents = len(payload)
         self.logger.debug("Sending %s events to splunk" % numberevents)
+        session_info_list = []
```
li-wu: is `session_info_list` used anywhere?

splunk_eventgen/lib/plugins/output/httpevent.py
```diff
@@ -182,50 +181,82 @@ def _sendHTTPEvents(self, payload):
...
+            target_url = '{}://{}:{}'.format(server_info.get('protocol', ''), server_info.get('address', ''), server_info.get('port', '')) 
+            if target_url in url:
+                if countdown_map[url] <= 0:
+                    del httpeventServers.get('servers')[i]
+                    self.logger.warning("Cannot reach {}. Removing from the server pool".format(url))
+                else:
+                    countdown_map[url] -= 1
+                    self.logger.debug("Cannot reach {}. Lowering countdown to {}".format(url, countdown_map[url]))
+
+        if isinstance(self.config.httpeventServers, str):
```
li-wu: `self.config.httpeventServers` => `httpeventServers ` ?

li-wu:
If we add the configuration `httpeventAllowFailureCount` , the corresponding spec and doc file should be updated too.

splunk_eventgen/lib/plugins/output/httpevent.py
```diff
@@ -182,50 +181,82 @@ def _sendHTTPEvents(self, payload):
...
                 self.logger.error("Failed sending events to url: %s  sourcetype: %s  size: %s" %
-                                  (url, self.lastsourcetype, payloadsize))
-                self.logger.debug(
-                    "Failed sending events to url: %s  headers: %s payload: %s" % (url, headers, payloadstring))
-                raise e
+                                  (url, self.lastsourcetype, len(payloadstring)))
+    
+    def reset_count(self, url):
+        try:
+            self.config.httpeventServersCountdownMap['url'] = self.httpeventAllowFailureCount
```
li-wu: the `url` is used as key here, not the actual string?

splunk_eventgen/lib/plugins/output/httpevent.py
```diff
@@ -267,24 +298,19 @@ def flush(self, q):
...
-                                "Server returned an error while sending, response code: %s" % response.status_code)
-                else:
+                if not self.config.httpeventWaitResponse:
                     self.logger.debug("Ignoring response from HTTP server, leaving httpevent outputter")
+                else:
+                    for session_info in self.active_session_info:
+                        url, session = session_info[0], session_info[1]
+                        try:
+                            response = session.result(3)
+                            self.reset_count(url)
```
li-wu: if reset the count each time for one specific url, the url seems will not be removed forever?

splunk_eventgen/lib/plugins/output/httpevent.py
arctan5x: great catch

splunk_eventgen/lib/plugins/output/httpevent.py
arctan5x: great catch

splunk_eventgen/lib/plugins/output/httpevent.py
arctan5x: I want to prevent a working server to be dropped because of some flakiness. I have seen that Splunk (without an ample size queue) returns error responses even though it is alive. Therefore, even after a series of failures, if we see it working again, we don't remove it.

splunk_eventgen/lib/plugins/output/httpevent.py
li-wu: so here you want keep type consistent for `self.config.httpeventServers`.  If it was `str`, keep `str` value and if it is `object`, keep `object` value.  why not apply this to `self._sample.httpeventServers`, previous value is `str` while `httpeventServers` is object and assigned to it.

splunk_eventgen/lib/plugins/output/httpevent.py
li-wu: make sense

PR #234, opened 2019-06-05
fix modinput init error

li-wu:


li-wu:
merge it since it is a hotfix before release 6.4.0

PR #236, opened 2019-06-06
Add deprecated info in doc for issue 225

li-wu:


docs/CONFIGURE.md
```diff
@@ -55,22 +54,22 @@ eventgen settings. Below is a list of those options with their descriptions:
     threading = thread | process
     * Configurable threading model. Process uses multiprocessing.Process in Python to get around issues with the GIL.
     * Defaults to thread
----
```
lephino: Why are we removing the line breaks or is this just that a linter hit it?

GordonWang: For github markdown render, an empty line is translated as an empty line. And tribble dashes are translated as a line separator.
https://gist.github.com/shaunlebron/746476e6e7a4d698b373

PR #238, opened 2019-06-11
enable test coverage in circleci

li-wu:


li-wu:
Pls hold the review util I make the htmlcov is successfully uploaded.

li-wu:
The pr is ready for review now. You can check the coverage report here: https://480-5445179-gh.circle-artifacts.com/0/home/circleci/project/htmlcov/index.html as an example.

Currently we only have 6% coverage.  
1. We did not enable unittest in the CI since it has lots of failures; 
2. Most functional tests do not have coverage report;

arctan5x:
@li-wu This is amazing. Now we can quantify our test coverage and set proper goals. 

Seems like you are running coverage on all levels of tests. Which level of test is the html report in the CircleCI artifact for? Is there a way to separate coverage for small, medium, and large and store the reports separately?

li-wu:
> @li-wu This is amazing. Now we can quantify our test coverage and set proper goals.
> 
> Seems like you are running coverage on all levels of tests. Which level of test is the html report in the CircleCI artifact for? Is there a way to separate coverage for small, medium, and large and store the reports separately?

The code coverage is an aggregation of small/medium/large test cases.  Since we currently do not have enough coverage for all the test case. Will separate if we do get enough coverage.

PR #239, opened 2019-06-12
Fix bug 230 and add test case

li-wu:
@lephino I removed the `self.workerQueue.unfinished_tasks > 0` and add corresponding test case for this. All functional tests passed. This might not be the final solution as we discussed in the issue. Will change based on your comments.

### DIFF
--- a/splunk_eventgen/eventgen_core.py
+++ b/splunk_eventgen/eventgen_core.py
@@ -548,8 +548,7 @@ class EventGenerator(object):
         :return:
         '''
         try:
-            while not self.sampleQueue.empty() or self.sampleQueue.unfinished_tasks > 0 or not self.workerQueue.empty(
-            ) or self.workerQueue.unfinished_tasks > 0:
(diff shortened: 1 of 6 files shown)


li-wu:
CI job failed due to required lib install failure. This will be fixed after pr #238 is merged.

splunk_eventgen/eventgen_core.py
```diff
@@ -548,8 +548,7 @@ def join_process(self):
         :return:
         '''
         try:
-            while not self.sampleQueue.empty() or self.sampleQueue.unfinished_tasks > 0 or not self.workerQueue.empty(
-            ) or self.workerQueue.unfinished_tasks > 0:
+            while not self.sampleQueue.empty() or self.sampleQueue.unfinished_tasks > 0 or not self.workerQueue.empty():
```
arctan5x: in threaded mode, I am concerned that the jobs might end prematurely. How about implement a check to see if the mode is threading and check if there are unfinished tasks?

PR #240, opened 2019-06-12
fix metrics output mode

Yangxulight:
Put metrics output plugin code to develop


splunk_eventgen/eventgen_core.py
```diff
@@ -346,6 +353,13 @@ def _setup_loggers(self, args=None, config=None):
             logger.handlers = []
             logger.addHandler(server_file_handler)
             logger.addHandler(console_handler)
+
+            # Configure httpeventout logger
+            logger = logging.getLogger('eventgen_httpeventout')
+            logger.setLevel(logging.INFO)
+            logger.propagate = False
+            logger.handlers = []
```
arctan5x: This line removes the logger handlers that were previously attached. Would you check if other handlers persist?

li-wu: The logger object is got by name `eventgen_httpeventout`, previous attached handlers seems not affected by this.

PR #246, opened 2019-06-26
Bug/issue245

technimad:
The regex which matches the token replacement configuration tries to match a minus sign, followed by either a value without a dot (integer) or a value with a dot (float). This match fails, because the current regex only matches a value with or without a dot (int) which could be preceded by a minus sign or a value with a dot (float). 
This subtile difference makes that this functionality doesn't work.

Arbitrarily used lastMileRTT to test the functionality.


arctan5x:
Hey thanks for contributing. For Splunk open source projects, we require contributors to fill out this form [https://www.splunk.com/goto/individualcontributions].

technimad:
Done. Filled out the form.

PR #264, opened 2019-07-09
Refactor the logging logic

li-wu:


li-wu:
This is a large PR. Main changes are:
1. Adopt `structlog` for better context awareness and debugging.
After import the `logger` for example `from logging_config import logger`, you could use the logger as a standard logger with additional context for example:
```
logger.info("Sample is loaded", name=self.sample.name, path=self.sample.path)
```
Result log msg should be:
```
2019-07-09 15:51:11 eventgen        INFO     MainProcess {'event': 'Logging Setup Complete.', 'name': 'sample_name', 'path': 'path_str'}
```
This would be efficient for debugging especially with Splunk. 

2. Remove duplicate code especially for setup_logger, remove logger from `__getstate__` and re-set the logger in `__setstate__`.

3. Remove `debugv` from the code base.

4. Replace `warn` with `warning` as `warn` is deprecated.

Things remain unchanged:
1. Logging settings such as logging level, logger name, logging file name etc.
2.Multiprocess logging logic. I keep this part logic unchanged until a better solution to avoid performance issue etc.
3. Ways to using the `logger`.

splunk_eventgen/eventgen_core.py
```diff
@@ -78,12 +56,11 @@ def __init__(self, args=None):
         self._generator_queue_size = getattr(self.args, 'generator_queue_size', 500)
         if self._generator_queue_size < 0:
             self._generator_queue_size = 0
-        self.logger.info("set generator queue size to %d", self._generator_queue_size)
+        self.logger.info("Set generator queue size to %d" % self._generator_queue_size)
```
lephino: we really should convert this to "".format, this isn't supported in py3

li-wu: Changed.

PR #265, opened 2019-07-10
Server revamp

arctan5x:
- Migrating the Nameko server to a new Flask server with new blueprint architecture.
- Fixing multiple bugs on endpoints.
- Standalone mode support
- No longer using RabbitMQ since we moved away from Nameko RPC. Going with HTTP calls now with healthchecks
- Test fixes
- Doc fixes


li-wu:
@arctan5x Could you briefly describe the target of the PR and solution?

PR #266, opened 2019-07-11
Server revamp

arctan5x:


splunk_eventgen/eventgen_api_server/constants.py
```diff
@@ -0,0 +1,14 @@
+DEBUG = True
```
li-wu: This file is not used and should be deleted( as well as related imports).

splunk_eventgen/__main__.py
```diff
@@ -338,6 +191,12 @@ def convert_verbosity_count_to_logging_level(verbosity):
     else:
         return logging.ERROR
 
+def gather_env_vars(args):
+    os_vars, env_vars = dict(os.environ), {}
+    env_vars["REDIS_HOST"] = os_vars.get("REDIS_HOST") if os_vars.get("REDIS_HOST") else args.redis_host
```
li-wu: Suggest `env_vars["REDIS_HOST"] = os.environ.get("REDIS_HOST", args.redis_host)`.

li-wu:
I saw a lots of coding format issue (spacing, unused imports, line length etc.) on `PyCharm`. Before we have `lint` in CI job, I think we should probably pay attention to it in each PR.

splunk_eventgen/eventgen_api_server/eventgen_server_api.py
```diff
@@ -0,0 +1,584 @@
...
+                return Response(json.dumps(self.get_conf()), mimetype='application/json', status=200)
+            except Exception as e:
+                self.logger.error(e)
+                return Response(INTERNAL_ERROR_RESPONSE, mimetype='application/json', status=500)
+
+        return bp
+
+    def get_index(self):
+        home_page = '''*** Eventgen WSGI ***\nHost: {0}\nEventgen Status: {1}\nEventgen Config file exists: {2}\nEventgen Config file path: {3}\nTotal volume: {4}\nWorker Queue Status: {5}\nSample Queue Status: {6}\nOutput Queue Status: {7}\n'''
+        status = get_status()
```
li-wu: Should be `status = self.get_status()` or it will throw `NameError`.

li-wu:
We should also change related docs to reflect the changes: http://splunk.github.io/eventgen/ARCHITECTURE.html

jmeixensperger:
> I saw a lots of coding format issue (spacing, unused imports, line length etc.) on `PyCharm`. Before we have `lint` in CI job, I think we should probably pay attention to it in each PR.

Not sure I agree here. Doing manual linting is very error-prone and it will slow down our development process.

li-wu:
> > I saw a lots of coding format issue (spacing, unused imports, line length etc.) on `PyCharm`. Before we have `lint` in CI job, I think we should probably pay attention to it in each PR.
> 
> Not sure I agree here. Doing manual linting is very error-prone and it will slow down our development process.

we do have `make format` command to do this **before** committing. You do not need to to it manually.

splunk_eventgen/eventgen_core.py
```diff
@@ -435,6 +438,8 @@ def _proc_worker_do_work(work_queue, logging_queue, config):
             except Empty:
                 stopping = genconfig['stopping']
             except Exception as e:
+                if self.force_stop:
```
li-wu: `_proc_worker_do_work` is a static method which do not have `self` context.

splunk_eventgen/eventgen_api_server/eventgen_core_object.py
```diff
@@ -0,0 +1,56 @@
...
+        args.backfill = None
+        args.count = None
+        args.devnull = False
+        args.disableOutputQueue = False
+        args.end = None
+        args.generators = None
+        args.interval = None
+        args.keepoutput = False
+        args.modinput = False
+        args.multiprocess = True
```
li-wu: it will use `multiprocess` by default?

rfaircloth-splunk:
There is something wrong with CI, while checks report passed on the PR 22 failures occurred if you inspect the job on circle

rfaircloth-splunk:
It looks like the docker image is redis based, normal practice with docker is "single thing" why are we not using redis in its own container and building eventgen as a separate container

dockerfiles/Dockerfile
```diff
@@ -10,17 +10,31 @@ RUN apk --no-cache upgrade && \
...
 	pip install --upgrade pip && \
 	rm -rf /tmp/* && \
-	rm -rf /var/cache/apk/*
+	rm -rf /var/cache/apk/* && \
+	ssh-keygen -f /etc/ssh/ssh_host_rsa_key -N '' -t rsa && \
+	mkdir -p /var/run/sshd && \
+	mkdir -p /root/.ssh && \
+	chmod 0700 /root/.ssh && \
+	passwd -u root && \
+	pip install git+git://github.com/esnme/ultrajson.git
```
li-wu: Does it necessary to install via git? we have `ujson` in `requirements.txt` already.

li-wu:
I push one commit to fix the failed test cases.

li-wu:
@arctan5x , either commit `be3ca13` or `804bd6f` introduced the three test cases failed: https://circleci.com/gh/splunk/eventgen/693

dockerfiles/Dockerfile
YifengMao: https://github.com/esnme/ultrajson/issues/326

li-wu: Thanks for the info.

li-wu:
> There is something wrong with CI, while checks report passed on the PR 22 failures occurred if you inspect the job on circle

CircleCI works like if you have a dedicated `test` step which returns none-zero code it will report `FAILED` status. Since the whole test cases are running in a container and the test flow always return zero status even it has failed test cases.  We may need to find a better to deal with this.

arctan5x:
At this point, all tests are fixed except for 1 failure which seems like a real bug, so I will investigate it tomorrow.

tests/large/test_eventgen_orchestration.py
```diff
@@ -282,7 +282,7 @@ def test_server_restart(self):
     def test_server_stop(self):
         r = requests.post("http://127.0.0.1:{}/stop".format(self.server_eventgen_webport))
         assert r.status_code == 200
-        assert "Eventgen is stopped" in r.content
+        assert "There is no Eventgen process running." in r.content
```
li-wu: Will `restart` operation restarts Eventgen process as previous configured?

.circleci/config.yml
```diff
@@ -16,6 +16,7 @@ jobs:
       - run:
           name: Run Tests
           command: |
+            set -e
```
li-wu: this is the default behavior?  https://circleci.com/docs/2.0/configuration-reference/#default-shell-options

dockerfiles/entrypoint.sh
```diff
@@ -1,11 +1,12 @@
...
 set -e
+/usr/sbin/sshd
 
 if [ "$#" = 0 ]; then
 	tail -F -n0 /etc/hosts && wait
 elif [ "$1" = "controller" ]; then
-	rabbitmq-server &
+	redis-server &
 	splunk_eventgen service --role controller &
 	tail -F -n0 /etc/hosts && wait
```
li-wu: Not for this pr. Why we need to read the `hosts` here?

docs/ARCHITECTURE.md
```diff
@@ -46,11 +46,16 @@ Given the complexity and the reimplementation of a number of features during ref
...
+This is a new feature included in version >= 6.0 Traditionally, it has been difficult to configure multiple Eventgen instances at the same time. The performance of a single Eventgen instance is often limited in its architecture and compute power of the host machine. We introduce Controller-Server architecture to support scalable Eventgen deployment.
 
-In Server-Controller Architecture, we may have one or multiple servers (Eventgen instances), one controller and one instance of RabbitMQ. It is ok to run RabbitMQ locally with the controller.
-The Servers and controller are communicating via RabbitMQ, where the controller has a capability to broadcast incoming requests to the servers and aggregate the results to output to the users.
+In this Architecture, we may have one or multiple servers (Eventgen instances), one controller and one instance of Redis (which lives inside of the same container as the container). You don't have to worry about setting up Redis yourself. Redis is used for communication between Controller and Servers.
```
li-wu: `which lives inside of the same container as the container` => `which lives inside of the same container as the controller`.

docs/SETUP.md
```diff
@@ -101,18 +101,15 @@ $ python -m splunk_eventgen -v generate tests/sample_eventgen_conf/replay/eventg
...
-8. **NOTE:** Running the controller and server on the same machine will cause port collisions for Eventgen web server. To mitigate this, you can tell the server to run on a separate port using `splunk_eventgen service --web-server-address 0.0.0.0:9501`
+##### Controller-Server Cluster ###
+
+Please follow these instructions to run an Eventgen cluster on your Docker environment:
+
+1. `make image`
+2. Create a Docker network: `docker network create --attchable --driver bridge eg_network`
+3. To set up a controller, run `docker run --network eg_network --name eg_controller  -d -p 6379:6379 -p 9500:9500 eventgen:latest controller`
+4. To set up a server, run `docker run --network eg_network --name eg_server -e REDIS_HOST=eg_controller -d -p 9501:9500 eventgen:latest server`
+* Note that REDIS_HOST needs to be a resolvable host address to the controoler. Also, --name should be used to differientiate a server from another.
```
li-wu: typo: `controoler` => `controller`.

dockerfiles/entrypoint.sh
arctan5x: etc/hosts don't mean anything, we are just tailing to keep the process running. We can tail any file whatsoever.

.circleci/config.yml
arctan5x: yeah, i added to see if it would fix the ci issue, but didn't work

tests/large/test_eventgen_orchestration.py
arctan5x: yes

dockerfiles/entrypoint.sh
li-wu: maybe `tail -f /dev/null` is better.

arctan5x: is there a difference between tailing /etc/hosts and /dev/null?

tests/large/test_eventgen_orchestration.py
li-wu: The the assertion should be `Eventgen is stopped` as previous test case restarts the server, right?

YifengMao:
tested the latest code, the process# issue is fixed.
but the performance on es bundle is likely 1/3 compares to 6.3.4. So the performance issue is not related to number of processes.

That is to say, the performance (80 processes in this code base) is 70% of 20 processes in 6.3.4 on the same hardware.

the performance issue is not related to the server change, because it can be reproduced in CLI mode.
and I saw plenty of logs from httpevent outputter:
`2019-07-24 02:13:18 eventgen        ERROR    Process-15 {'event': 'failure outputting event, does not contain _raw'}
2019-07-24 02:13:18 eventgen        ERROR    Process-15 {'event': 'failure outputting event, does not contain _raw'}
2019-07-24 02:13:18 eventgen        ERROR    Process-15 {'event': 'failure outputting event, does not contain _raw'}
2019-07-24 02:13:18 eventgen        ERROR    Process-15 {'event': 'failure outputting event, does not contain _raw'}
2019-07-24 02:13:18 eventgen        ERROR    Process-15 {'event': 'failure outputting event, does not contain _raw'}
2019-07-24 02:13:18 eventgen        ERROR    Process-15 {'event': 'failure outputting event, does not contain _raw'}
2019-07-24 02:13:18 eventgen        ERROR    Process-15 {'event': 'failure outputting event, does not contain _raw'}
2019-07-24 02:13:18 eventgen        ERROR    Process-15 {'event': 'failure outputting event, does not contain _raw'}`

it might be the root cause of why we have such critical throughput loss

arctan5x:
@YifengMao I think it is a good point. Let's check the http outputter.

splunk_eventgen/eventgen_api_server/api_blueprint.py
```diff
@@ -0,0 +1,6 @@
+class ApiBlueprint():
```
jmeixensperger: Any purpose for this class?

li-wu:
@arctan5x you mentioned in last sync up meeting that "Missing the logic where Eventgen Controller needs to know dropped servers", does it covered now?

splunk_eventgen/eventgen_core.py
```diff
@@ -507,3 +533,16 @@ def check_done(self):
         :return: if eventgen jobs are finished, return True else False
         '''
         return self.sampleQueue.empty() and self.sampleQueue.unfinished_tasks <= 0 and self.workerQueue.empty() and self.workerQueue.unfinished_tasks <= 0
+
+    def kill_processes(self):
+        if self.args.multiprocess:
```
li-wu: how about `multiprocess` is assigned in `conf` instead of in `args`?

splunk_eventgen/eventgen_core.py
```diff
@@ -61,6 +64,8 @@ def __init__(self, args=None):
         if self.args and 'configfile' in self.args and self.args.configfile:
             self._load_config(self.args.configfile, args=args)
 
+        self.workerPool = []
```
li-wu: duplicate with line 241?

splunk_eventgen/eventgen_core.py
```diff
@@ -6,6 +6,8 @@
 import os
 import sys
 import time
+import subprocess
```
li-wu: not used import?

arctan5x:
@li-wu We don't have it still, but that feature only affects /index endpoint which is very low impact and no functional effect. So going to release this then add it as a backlog.

splunk_eventgen/eventgen_api_server/eventgen_core_object.py
arctan5x: I am on a debate about this. I would rather have Eventgen server to work in one mode to keep the implementation consistent. It is just too much hassle to support both threading and multiprocessing. In this situation, I would support multiprocessing by default, because the use case is more wide.

PR #272, opened 2019-07-20
Docker file enhancements

rfaircloth-splunk:
Remove ssh and developer cruft
Use multi layer build to package eventgen
TODO: swtich to py2 base image and use redis in a dedicated container


dockerfiles/Dockerfile
```diff
@@ -10,29 +26,18 @@ RUN apk --no-cache upgrade && \
...
-	acl \
-	g++ \
-	git \
-	curl && \
+	curl \
+	bash && \
 	pip install --upgrade pip && \
 	rm -rf /tmp/* && \
-	rm -rf /var/cache/apk/* && \
-	ssh-keygen -f /etc/ssh/ssh_host_rsa_key -N '' -t rsa && \
```
arctan5x: We need SSH for lots of internal use cases through microservices.

dockerfiles/entrypoint.sh
```diff
@@ -1,20 +1,16 @@
...
 if [ "$#" = 0 ]; then
 	tail -F -n0 /etc/hosts && wait
 elif [ "$1" = "controller" ]; then
 	redis-server &
-	splunk_eventgen service --role controller &
-	tail -F -n0 /etc/hosts && wait
+	splunk_eventgen service --role controller
 elif [ "$1" = "server" ]; then
-	splunk_eventgen service --role server &
-	tail -F -n0 /etc/hosts && wait
```
arctan5x: Because of the use cases internally, we would rather prefer to keep the container running despite the Eventgen process dies.

.circleci/config.yml
```diff
@@ -24,10 +24,39 @@ jobs:
           path: /home/circleci/project/tests/test-reports
       - store_artifacts:
           path: /home/circleci/project/htmlcov
+  build_docker:
+    environment:
+      IMAGE_ACCOUNT: rfaircloth
```
arctan5x: I do not think it would be suitable to add your Docker build process in the EventgenCI. I would recommend creating a separate CI pipeline that pulls from this repo for your personal use case.

rfaircloth-splunk:
Agree I had that on my branch for  my needs I forgot the PR would update

arctan5x:
Hi Ryan, would this PR be still valid? If not may I close it?

rfaircloth-splunk:
Close it I will re-submit

PR #274, opened 2019-07-22
Update docs and license

li-wu:
* Fix modular input functional test case on CI
* Add structlog license
* Fix doc format


li-wu:
One test case failed but passed on my local env. I am figuring why it failed on CI.

li-wu:
> One test case failed but passed on my local env. I am figuring why it failed on CI.

Now the PR is ready for review now.

PR #287, opened 2019-08-14
Fix bug 286 random token replacement

li-wu:
Replace with `random.SystemRandom`.

### DIFF
--- a/splunk_eventgen/lib/eventgentoken.py
+++ b/splunk_eventgen/lib/eventgentoken.py
@@ -350,7 +350,7 @@ class Token(object):
                 except:
                     logger.error("Could not parse json for '%s' in sample '%s'" % (listMatch.group(1), s.name))
                     return old
-                return random.choice(value)
+                return random.SystemRandom().choice(value)
 
             else:
                 logger.error("Unknown replacement value '%s' for replacementType '%s'; will not replace" %
--- a/splunk_eventgen/lib/plugins/generator/perdayvolumegenerator.py
+++ b/splunk_eventgen/lib/plugins/generator/perdayvolumegenerator.py
@@ -31,17 +31,15 @@ class PerDayVolumeGenerator(GeneratorPlugin):
 
         # Create a counter for the current byte size of the read in samples
         currentSize = 0
-        # Replace event tokens before calculating the size of the event
-        updated_sample_dict = GeneratorPlugin.replace_tokens(self, self._sample.sampleDict, earliest, latest)
 
         # If we're random, fill random events from sampleDict into eventsDict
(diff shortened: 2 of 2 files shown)


szhong26:
It looks like this is still not correct. Here is a sample output for the sample CPUTime.perfmon
[eventgen_data_output.txt](https://github.com/splunk/eventgen/files/3502456/eventgen_data_output.txt)

splunk_eventgen/lib/eventgentoken.py
```diff
@@ -350,7 +350,7 @@ def _getReplacement(self, old=None, earliestTime=None, latestTime=None, s=None,
                 except:
                     logger.error("Could not parse json for '%s' in sample '%s'" % (listMatch.group(1), s.name))
                     return old
-                return random.choice(value)
+                return random.SystemRandom().choice(value)
```
li-wu: Use `SystemRandom` class which uses the system function `os.urandom()` to generate random numbers from sources provided by the operating system.

splunk_eventgen/lib/plugins/generator/perdayvolumegenerator.py
```diff
@@ -31,17 +31,15 @@ def gen(self, count, earliest, latest, samplename=None):
 
         # Create a counter for the current byte size of the read in samples
         currentSize = 0
-        # Replace event tokens before calculating the size of the event
-        updated_sample_dict = GeneratorPlugin.replace_tokens(self, self._sample.sampleDict, earliest, latest)
```
li-wu: Use the raw sample to calculate the size of event and replace the token later in `build_events`.

splunk_eventgen/lib/plugins/generator/perdayvolumegenerator.py
```diff
@@ -63,28 +61,28 @@ def gen(self, count, earliest, latest, samplename=None):
...
                 sizeremaining = size - currentreadsize
-                targetlinesize = len(updated_sample_dict[targetline]['_raw'])
+                targetlinesize = len(self._sample.sampleDict[targetline]['_raw'])
                 if size < targetlinesize:
                     logger.error(
                         "Size is too small for sample {}. We need {} bytes but size of one event is {} bytes.".format(
                             self._sample.name, size, targetlinesize))
                     break
-                if targetlinesize <= sizeremaining or targetlinesize * .9 <= sizeremaining:
+                if targetlinesize <= sizeremaining:
```
li-wu: Remove redundant logic here.

li-wu:
Per feedback from @szhong26 ,  this pr now can get random value replacement for token.

PR #288, opened 2019-08-14
[global] perDayVolume

jmeixensperger:
We explicitly exclude this stanza from our stanza count when calculating perDayVolume, but we assign the setting the stanza anyways. This results in an inaccurate total perDayVolume when we append the values of each stanza.

My fix will prevent us from assigning the setting to the stanza to mirror how we count each stanza.

### DIFF
--- a/splunk_eventgen/eventgen_api_server/eventgen_server_api.py
+++ b/splunk_eventgen/eventgen_api_server/eventgen_server_api.py
@@ -371,7 +371,7 @@ class EventgenServerAPI():
                 stanza_num -= 1
             divided_volume = float(target_volume) / stanza_num
             for stanza, kv_pair in conf_dict.iteritems():
-                if isinstance(kv_pair, dict) and 'global' not in stanza and '.*' not in stanza:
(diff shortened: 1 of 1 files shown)


li-wu:
Is there an issue for more detail info of this PR? @jmeixensperger

jmeixensperger:
> Is there an issue for more detail info of this PR? @jmeixensperger

@li-wu No, this was found by one of our interns and affected their project.

splunk_eventgen/eventgen_api_server/eventgen_server_api.py
```diff
@@ -371,7 +371,7 @@ def set_volume(self, target_volume):
                 stanza_num -= 1
             divided_volume = float(target_volume) / stanza_num
             for stanza, kv_pair in conf_dict.iteritems():
-                if isinstance(kv_pair, dict) and stanza != '.*' not in stanza:
+                if isinstance(kv_pair, dict) and 'global' not in stanza and '.*' not in stanza:
```
li-wu: maybe we should use `stanza != global and '.*' not in stanza` to avoid sample name with `global` in it for example "[test_global_sample]".

PR #291, opened 2019-08-26
Py3 upgrade

li-wu:
Python3 support. This is a huge pr with a lot of changes and I will comment on the key changes once all test cases are passed.


splunk_eventgen/eventgen_core.py
```diff
@@ -300,8 +288,8 @@ def _proc_worker_do_work(work_queue, logging_queue, config):
         root = logging.getLogger()
         root.setLevel(logging.DEBUG)
         if logging_queue is not None:
-            # TODO https://github.com/splunk/eventgen/issues/217
-            qh = logutils.queue.QueueHandler(logging_queue)
```
li-wu: Remove `logutils` from source code and replace `QueueHandler` with `logging.handlers.QueueHandler`.

requirements.txt
```diff
@@ -2,15 +2,14 @@ pytest==4.6.4
...
+docker==3.7.3
 pyOpenSSL
 lxml==4.3.4
 pytest-mock>=1.10.4
 boto3
 requests>=2.18.4
 requests[security]
 logutils>=0.3.4.1
-futures>=3.0.5
 ujson>=1.35
```
li-wu: `futures` module have become a built-in module in `Python3`.

splunk_eventgen/eventgen_core.py
```diff
@@ -1,33 +1,21 @@
...
-# Since i'm including a new library but external sources may not have access to pip (like splunk embeded), I need to
-# be able to load this library directly from src if it's not installed.
-try:
-    import logutils
-    import logutils.handlers
-except ImportError:
-    path_prepend = os.path.join(os.path.dirname(os.path.abspath(__file__)), 'lib', 'logutils_src')
-    sys.path.append(path_prepend)
-    import logutils
-    import logutils.queue
```
li-wu: Change relative imports to absolute imports and remove all related path inserts.

splunk_eventgen/lib/eventgensamples.py
```diff
@@ -302,61 +301,52 @@ def latestTime(self):
...
         return self.now(utcnow=True)
 
-    def _openSampleFile(self):
-        logger.debug("Opening sample '%s' in app '%s'" % (self.name, self.app))
-        self._sampleFH = open(self.filePath, 'rU')
-
-    def _closeSampleFile(self):
-        logger.debug("Closing sample '%s' in app '%s'" % (self.name, self.app))
-        self._sampleFH.close()
-
```
li-wu: Remove the `self._sampleFH` which is a file handler and can not be pickled. This will cause `PickleError` in Python3.

splunk_eventgen/lib/plugins/generator/jinja.py
```diff
@@ -111,7 +109,7 @@ def parse(self, parser):
         target_var_name = {"time_now": "time_now", "time_slice": "time_target"}
         tag = parser.stream.current.value
         name_base = target_var_name[tag]
-        lineno = parser.stream.next().lineno
+        lineno = next(parser.stream).lineno
```
li-wu: `steam.next()` is removed and use `next` instead.

docs/SETUP.md
```diff
@@ -149,7 +149,12 @@ $ cd ${SPLUNK_HOME}/etc/apps
 ```
 
 ### First Run
-If SA-Eventgen App is correctly installed, there is no additional configuration required. SA-Eventgen app will automatically identify with any apps with eventgen.conf.
+If your `SA-Eventgen` is 7.0+, you need to add a new parameter under `[general]` stanza in `$SPLUNK_HOME/etc/system/local/server.conf`:
+```
+[general]
+python.version = python3
+```
+Then `SA-Eventgen` app will automatically identify with any apps with `eventgen.conf`.
```
li-wu: Currently the latest splunk build seems to have bugs when place this in to `inputs.conf`.

splunk_eventgen/__main__.py
```diff
@@ -1,25 +1,29 @@
-'''
```
GordonWang: we'd better consult splunk legal team to come up with a standard file headers for our eventgen project code.
Usually it describes the copy right or any legal concerns.

tests/large/test_mode_replay.py
```diff
@@ -69,7 +69,7 @@ def test_mode_replay_timemultiple(eventgen_test_helper):
         event_datetime = datetime.strptime(result.group(), "%Y-%m-%d %H:%M:%S")
         delter_seconds = (event_datetime - current_datetime).total_seconds()
         # assert the event time is after (now - earliest) time
-        assert delter_seconds < 12
+        assert delter_seconds < 14
```
jmeixensperger: Can we remove these hard-coded values from our tests? In my opinion, they should calculated from the configurations/samples in case we change the tests later.

docs/SETUP.md
li-wu: This bug is fixed in core now.

tests/large/test_mode_replay.py
li-wu: Sometimes this will not work. For example, the time interval got from sample file for above example is 12 but the actual `delter_seconds` could be 12.323 and cause the test failed.

li-wu:
This pull request is ready for review now. Thanks.

splunk_eventgen/__main__.py
li-wu: I updated some outdated info for this pr.

tests/large/test_mode_replay.py
jmeixensperger: We can address this in another PR, but I'm curious why 14? For larger time spans, do we also expect `delter_seconds + 2`? Or do we expect that the actual result can be a certain percentage higher than what we specify in the sample? I would like to define this more clearly so that we know which test failures are valid or not.

splunk_eventgen/default/eventgen.conf
```diff
@@ -1,4 +1,3 @@
-# Copyright (C) 2005-2015 Splunk Inc. All Rights Reserved.
```
GordonWang: Why do we remove this copyright line?

splunk_eventgen/eventgen_core.py
```diff
@@ -382,11 +368,16 @@ def _initializePlugins(self, dirname, plugins, plugintype, name=None):
                     try:
                         # Import the module
                         # module = imp.load_source(base, filename)
+
                         mod_name, mod_path, mod_desc = imp.find_module(base, [dirname])
                         # TODO: Probably need to adjust module.load() to be added later so this can be pickled.
                         module = imp.load_module(base, mod_name, mod_path, mod_desc)
                         plugin = module.load()
 
+                        # spec = importlib.util.spec_from_file_location(base, filename)
```
GordonWang: can we remove these commented code?

splunk_eventgen/default/eventgen.conf
li-wu: It is out of data and the copyright info is located in some file but not others. So I just removed it.

PR #296, opened 2019-09-11
Add syslogAddHeader config directive

patriknordlen:
This PR adds functionality to the syslogoutput plugin that if enabled makes the plugin output an RFC 3164 compliant header at the beginning of each message, consisting of a timestamp and a hostname. The value of the hostname is taken from the host value defined for the sample.

The use-case is in situations where eventgen is set to output events as syslog but one wants the receiving syslog server to be able to use the sample's defined host value as originating hostname. Without the initial header this is not possible and events are sent as-is, in which case the receiving syslog server has to use the hostname of the host that eventgen is running on instead. Upon adding this header, eventgen will emulate being an intermediate syslog forwarder, adding the "original" hostname at the start of the message.

### DIFF
--- a/splunk_eventgen/lib/eventgenconfig.py
+++ b/splunk_eventgen/lib/eventgenconfig.py
@@ -88,6 +88,7 @@ class Config(object):
                     'minuteOfHourRate', 'timezone', 'dayOfMonthRate', 'monthOfYearRate', 'perDayVolume',
                     'outputWorkers', 'generator', 'rater', 'generatorWorkers', 'timeField', 'sampleDir', 'threading',
                     'profiler', 'maxIntervalsBeforeFlush', 'maxQueueLength', 'splunkMethod', 'splunkPort',
+                    'syslogDestinationHost', 'syslogDestinationPort', 'syslogAddHeader',
                     'verbosity', 'useOutputQueue', 'seed','end', 'autotimestamps', 'autotimestamp', 'httpeventWaitResponse',
                     'outputCounter', 'sequentialTimestamp', 'extendIndexes', 'disableLoggingQueue']
     _validTokenTypes = {'token': 0, 'replacementType': 1, 'replacement': 2}
@@ -99,7 +100,7 @@ class Config(object):
     _floatSettings = ['randomizeCount', 'delay', 'timeMultiple']
     _boolSettings = [
         'disabled', 'randomizeEvents', 'bundlelines', 'profiler', 'useOutputQueue', 'autotimestamp',
-        'httpeventWaitResponse', 'outputCounter', 'sequentialTimestamp', 'disableLoggingQueue']
+        'httpeventWaitResponse', 'outputCounter', 'sequentialTimestamp', 'disableLoggingQueue', 'syslogAddHost']
     _jsonSettings = [
         'hourOfDayRate', 'dayOfWeekRate', 'minuteOfHourRate', 'dayOfMonthRate', 'monthOfYearRate', 'autotimestamps']
     _defaultableSettings = [
--- a/splunk_eventgen/lib/eventgenconfig.py
+++ b/splunk_eventgen/lib/eventgenconfig.py
@@ -100,7 +100,7 @@ class Config(object):
     _floatSettings = ['randomizeCount', 'delay', 'timeMultiple']
     _boolSettings = [
         'disabled', 'randomizeEvents', 'bundlelines', 'profiler', 'useOutputQueue', 'autotimestamp',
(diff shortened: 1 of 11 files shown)


li-wu:
Do we have an issue to track this?

patriknordlen:
I didn't create an issue, I saw the need for this and implemented the solution instead of filing an issue, but I can create one if it's preferred for tracking purposes etc!

patriknordlen:
@li-wu should I create an issue to match the PR against?

li-wu:
> @li-wu should I create an issue to match the PR against?

Yes please thanks.

patriknordlen:
@li-wu Done! https://github.com/splunk/eventgen/issues/299

splunk_eventgen/lib/eventgenconfig.py
```diff
@@ -99,7 +100,7 @@ class Config(object):
     _floatSettings = ['randomizeCount', 'delay', 'timeMultiple']
     _boolSettings = [
         'disabled', 'randomizeEvents', 'bundlelines', 'profiler', 'useOutputQueue', 'autotimestamp',
-        'httpeventWaitResponse', 'outputCounter', 'sequentialTimestamp', 'disableLoggingQueue']
+        'httpeventWaitResponse', 'outputCounter', 'sequentialTimestamp', 'disableLoggingQueue', 'syslogAddHost']
```
li-wu: This should be `syslogAddHeader`?

splunk_eventgen/lib/plugins/output/syslogout.py
```diff
@@ -8,16 +8,24 @@
 # Dict of flags to gate adding the syslogHandler only once to the given singleton logger
 loggerInitialized = {}
 
+class HostFilter(logging.Filter):
+    def __init__(self, host):
+        self.host = host
+
+    def filter(self, record):
```
li-wu: This filter always returns `True`.

patriknordlen: Correct, this is as intended - the filter should never return False, because all events should be sent. The purpose of the filter is not to remove messages from the output stream, it is only to add the host field so it's available in the formatter further down.

splunk_eventgen/lib/eventgenconfig.py
patriknordlen: Indeed it should! Will push a fix.

tests/sample_eventgen_conf/medium_test/eventgen.conf.syslogoutput
```diff
@@ -8,4 +8,4 @@ end = 1
 outputMode = syslogout
 syslogDestinationHost = 127.0.0.1
 syslogDestinationPort = 9999
-
+syslogAddHeader = false
```
GordonWang: can we add one more test case to address this feature?
by default it is false. 
We need one more case to test if syslogAddHeader=true

splunk_eventgen/lib/plugins/output/syslogout.py
```diff
@@ -26,6 +34,8 @@ def __init__(self, sample, output_counter=None):
 
         loggerName = 'syslog' + sample.name
         self._l = logging.getLogger(loggerName)
+        if syslogAddHeader:
+            self._l.addFilter(HostFilter(host=sample.host))
```
GordonWang: please update the host setting in configuration docs in [eventgen.conf.spec](https://github.com/splunk/eventgen/blob/c48874703a09756847b3de91eaac2e6b49c8a381/splunk_eventgen/splunk_app/README/eventgen.conf.spec#L223)
Because now, host is not only used in splunkstream mode.

splunk_eventgen/lib/plugins/output/syslogout.py
GordonWang: please add this description as comment in HostFilter::filter method.
It would be much easier for other people to understand the purpose of this filter.

GordonWang:
Thanks for this PR. Please update the test case to address this feature.

tests/sample_eventgen_conf/medium_test/eventgen.conf.syslogoutput
patriknordlen: I added a new separate test for this - considered integrating it into existing one but thinking it might be confusing.

patriknordlen:
> Thanks for this PR. Please update the test case to address this feature.

Thanks for the review. I pushed changes based on your comment - please review and let me know if anything should be done differently.

PR #301, opened 2019-09-26
Using multiprocess pool to address the OOM issue

li-wu:
Previously, when using `multiprocess` mode with 20 `generatorWorkers`, the process will be killed as `Out of memory` error.

After adopting `pool` with `maxtasksperchild`, the processes are now down for OOM after more than 24 hours. 

Data ingestion with `perdayvolume=600`:
image

Process status and memory usage:
image

### DIFF
--- a/splunk_eventgen/eventgen_core.py
+++ b/splunk_eventgen/eventgen_core.py
@@ -6,9 +6,9 @@ import logging.config
 import os
 import sys
 import time
-import signal
 from Queue import Empty, Queue
 from threading import Thread
+import multiprocessing
 
 from lib.eventgenconfig import Config
 from lib.eventgenexceptions import PluginNotLoaded
@@ -140,7 +140,7 @@ class EventGenerator(object):
         # APPPERF-263: be greedy when scanning plugin dir (eat all the pys)
         self._initializePlugins(plugindir, pluginsdict, plugintype)
 
-    def _setup_pools(self, generator_worker_count):
+    def _setup_pools(self, generator_worker_count=20):
         '''
         This method is an internal method called on init to generate pools needed for processing.
 
@@ -150,7 +150,8 @@ class EventGenerator(object):
         self._create_generator_pool()
         self._create_timer_threadpool()
         self._create_output_threadpool()
-        self._create_generator_workers(generator_worker_count)
+        if self.args.multiprocess:
+            self.pool = multiprocessing.Pool(generator_worker_count, maxtasksperchild=1000000)
 
     def _create_timer_threadpool(self, threadcount=100):
         '''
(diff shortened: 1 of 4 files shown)


splunk_eventgen/lib/eventgenoutput.py
```diff
@@ -98,6 +98,7 @@ def flush(self, endOfInterval=False):
             outputer = self.outputPlugin(self._sample, self.output_counter)
             outputer.updateConfig(self.config)
             outputer.set_events(q)
+            del q
```
GordonWang: why do we need to to delete `q` and `outputer` explicitly?

splunk_eventgen/eventgen_core.py
```diff
@@ -426,8 +383,12 @@ def start(self, join_after_start=True):
                 self.logger.info("Creating timer object for sample '%s' in app '%s'" % (s.name, s.app))
                 # This is where the timer is finally sent to a queue to be processed.  Needs to move to this object.
                 try:
-                    t = Timer(1.0, sample=s, config=self.config, genqueue=self.workerQueue,
-                              outputqueue=self.outputQueue, loggingqueue=self.loggingQueue)
+                    if self.args.multiprocess:
+                        t = Timer(0.1, sample=s, config=self.config, genqueue=self.workerQueue,
```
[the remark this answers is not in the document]
li-wu: It is typo. Fixed it.

splunk_eventgen/lib/eventgenoutput.py
li-wu: Previously I tried this to check if memory usage will downgrade if explicitly delete. It is better but we still got OOM error. I still keep this change.

li-wu:
I will check why the modular input test case failed.

li-wu:
Running more than three days with stable output and without process down. 

image

li-wu:
> I will check why the modular input test case failed.

All tests are passed now.

splunk_eventgen/eventgen_core.py
```diff
@@ -234,22 +238,6 @@ def _create_generator_pool(self, workercount=20):
                     worker.setDaemon(True)
                     worker.start()
 
-    def _create_generator_workers(self, workercount=20):
```
lephino: The original intent of this other structure was to allow the main eventgen program to receive config updates, change and dynamically recreate and allocate the pools.  The new style here will lock on the join's and will then wait for the pool.  Are we sure we want to go that route?  If you want to change config, you have to restart eventgen?  If we do, I think this is fine, however the "max_tasks" fix is really more about the queues getting backed up and filling up due to eventgen not being fast enough, and I think this is overkill for the fix.

li-wu: For `multiprocess` mode, the config will only load once after starting EventGen. (Correct me if I am wrong)

And by leveraging `max_tasks`, we leave the memory management task to the process pool. I have also tried other ways such as getting the queue larger, explicitly delete large object after usage which are not working to fix the OOM issue. It also takes long time to verify each try and test. Maybe this is not the best solution but it solves the issue for us to release the last version of Py2.

lephino: Yeah the original method that was created was to have the config object be able to have a "max workers" and to be able to update the worker count and then recreate all the pools and workers dynamical.  I'm fine if we kill off that logic, that's more of a lets make the call in our next meeting

li-wu: > the original method that was created was to have the config object be able to have a "max workers" and to be able to update the worker count and then recreate all the pools and workers dynamical.
This is the initial thought right? The currently logic is not what you said above. 

Then let's ship this and discuss it on our next meeting.

PR #306, opened 2019-10-10
variable assignment fix

arctan5x:


jmeixensperger:
Why is this needed? Can you give a description?

PR #308, opened 2019-10-11
Add Healthcheck Endpoint to Eventgen

Devmalion:
Adding a healthcheck endpoint to determine controller and server connections to Redis
The controller will periodically (every 30 minutes) hit the endpoint and evaluate the connection statuses.


li-wu:
The CI seems failed.

PR #317, opened 2019-10-17
test the makefile exit

GordonWang:
make the exit code as non zero when test case fails.
Fail fast when there are test case failure. Do not run large tests if there are failures in medium tests.


GordonWang:
guys, the PR is ready. Please review. 

I changed the test case to verify this change. 
https://circleci.com/gh/splunk/eventgen/903

Thanks.

PR #326, opened 2019-10-23
cleanup memory

arctan5x:


splunk_eventgen/lib/outputplugin.py
```diff
@@ -37,11 +37,6 @@ def run(self):
         if self.output_counter is not None:
             self.output_counter.collect(len(self.events), sum([len(e['_raw']) for e in self.events]))
         self.events = None
-        self._output_end()
-
-    def _output_end(self):
-        pass
-
```
Devmalion: I don't believe we should remove these lines. `CounterOutputPlugin` extends off the base `OutputPlugin` and overrides the _output_end method. Thus at the end of run we want to run `self._output_end()` to keep its expected behaviour.

PR #335, opened 2019-10-31
add release automation script

GordonWang:
As the title.


GordonWang:
The output of this automation script is like the diff in this PR
https://github.com/splunk/eventgen/pull/337

PR #336, opened 2019-10-31
Fix oom caused by ujson

li-wu:
After a bunch of testing on the OOM issue. I found this is caused by the `ujson` package which is needed by `conduct2` for perf testing. In this pr, I removed the installation step for `ujson` via git source code and I kicked two round of tests(one standalone and one triggered by perf test case) based on the new built docker image and the OOM issue is gone now.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -39,10 +39,10 @@ test_helper:
 	docker cp . ${EVENTGEN_TEST_IMAGE}:$(shell pwd)
 
 	@echo 'Verifying contents of pip.conf'
-	docker exec -i ${EVENTGEN_TEST_IMAGE} /bin/sh -c "cd $(shell pwd); pip install dist/splunk_eventgen*.tar.gz"
+	docker exec -i ${EVENTGEN_TEST_IMAGE} /bin/sh -c "cd $(shell pwd); pip3 install dist/splunk_eventgen*.tar.gz"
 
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -39,10 +39,10 @@ test_helper:
 	docker cp . ${EVENTGEN_TEST_IMAGE}:$(shell pwd)
 
 	@echo 'Verifying contents of pip.conf'
-	docker exec -i ${EVENTGEN_TEST_IMAGE} /bin/sh -c "cd $(shell pwd); pip install dist/splunk_eventgen*.tar.gz"
+	docker exec -i ${EVENTGEN_TEST_IMAGE} /bin/sh -c "cd $(shell pwd); pip3 install dist/splunk_eventgen*.tar.gz"
 
 	@echo 'Installing test requirements'
-	docker exec -i ${EVENTGEN_TEST_IMAGE} /bin/sh -c "pip install --upgrade pip;pip install -r $(shell pwd)/requirements.txt"
+	docker exec -i ${EVENTGEN_TEST_IMAGE} /bin/sh -c "pip3 install --upgrade pip;pip3 install -r $(shell pwd)/requirements.txt;pip3 install git+https://github.com/esnme/ultrajson.git"
```
li-wu: https://github.com/esnme/ultrajson/issues/326

PR #346, opened 2019-11-08
add throughput summary and fix timmer blocked by  end=3 count=0 settings

Yangxulight:
Here is what I did in this PR:
1. Add Throughput summary when we call "/status" from a controller. (eventgen_controller_api.py)
2. When we have a sample with settings of "count=0, end=3", the timer for that sample will blocked and only log "There is no data to be generated in worker...". because the execution is always 0. We should increase the execution so that this sample timer can finished the job after 3 execution.

### DIFF
--- a/splunk_eventgen/eventgen_api_server/eventgen_controller_api.py
+++ b/splunk_eventgen/eventgen_api_server/eventgen_controller_api.py
@@ -17,20 +17,20 @@ class EventgenControllerAPI:
         self.interval = 0.001
 
         self.server_responses = {}
-    
+
     def get_blueprint(self):
         return self.bp
-    
+
     def __create_blueprint(self):
         bp = Blueprint('api', __name__)
-        
+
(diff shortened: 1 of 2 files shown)


arctan5x:
Would you add a functional test for throughput summary?

li-wu:
> Would you add a functional test for throughput summary?

+1. Especially for the end = 3 and count = 0 setting scenario. Thanks, Ryan.

Yangxulight:
It seems that the CircleCI was blocked by my test code(ec9192c). Do you guys have some ideas about the reason?

li-wu:
> It seems that the CircleCI was blocked by my test code([ec9192c](https://github.com/splunk/eventgen/commit/ec9192c47d1742dec36c9d86446b8f811b81145c)). Do you guys have some ideas about the reason?

Could you place the test logic in another test function since you need to set the conf and start eventgen and validate the logic?

PR #353, opened 2019-11-28
Add multithread support

li-wu:


arctan5x:
Hi, we also have a logic in setup_http function to update threading option to process.

splunk_eventgen/eventgen_api_server/eventgen_core_object.py
```diff
@@ -43,7 +43,7 @@ def _create_args(self):
         args.interval = None
         args.keepoutput = False
         args.modinput = False
-        args.multiprocess = True
+        args.multiprocess = False if kargs.get("mutithread") else True
```
arctan5x: typo

li-wu: Fixed.

li-wu:
> setup_http

you mean the logic here?
https://github.com/splunk/eventgen/blob/develop/splunk_eventgen/eventgen_api_server/eventgen_server_api.py#L536

arctan5x:
@li-wu Yes. That is correct.

li-wu:
@arctan5x I walked through the function `setup_http` and it seems the `threading` is set to `process` and no way to change it to `thread`.  Any thing missing?

li-wu:
@arctan5x could you help confirm my previous comment? Thanks.

arctan5x:
What do you mean by that there is no way to change it to "thread"?

arctan5x:
Seems like that there is no dynamic way to change threading mode while the server is running. Since users don't control how server starts (only able to control threading through args/evn vars), how are users using this change?

li-wu:
> What do you mean by that there is no way to change it to "thread"?

I mean there is no logic to change the mode to thread mode in the function `setup_http `.

li-wu:
> Seems like that there is no dynamic way to change threading mode while the server is running. Since users don't control how server starts (only able to control threading through args/evn vars), how are users using this change?

If users like perf team are using the docker, that might be a problem since they are using /start REST endpoint to start the service. Any good suggestion?

Yangxulight:
It's ok for our perf team. Actually, we only need to make sure the eventgen server mode using multithread by default. I think we can change the `entrypoint.sh` with `--multithread` parameter on server mode when we want to upload a new image.

arctan5x:
If perf team is going to change entrypoint.sh, I would say this PR would do its job. One thing I would note is that /setup changes threading to multiprocess. So please note that when you use.

PR #363, opened 2020-03-13
Correct python2/3 issue

rfaircloth-splunk:
When sending via tcpout or udpout convert the msg to bytes as this was causing an exception


li-wu:
Thanks for contributing this PR. The CI failed for the changes, could you give more info about the exception?

Issue #364, opened 2020-03-20
[BUG] No events generated with regex in stanza


jmeixensperger:
**Describe the bug**
No events are generated with my stanza that includes a regex pattern.

**To Reproduce**
use attached conf/samples

**Expected behavior**
Events generated from multiple files

**Actual behavior**
No events

**Sample files and eventgen.conf file**
eventgen.conf:
[file\d+\\.txt]
mode = sample
outputMode = stdout
end = 1

**Do you run eventgen with SA-eventgen?**
No

PR #365, opened 2020-03-20
regex stanza bugfix

jmeixensperger:
Addresses #364


li-wu:
Seems 12 test cases failed by this PR. Could you take a further check, Jack? Thanks.

jmeixensperger:
@li-wu @GordonWang I'm having issues getting the tests to run locally. Let's make sure they are working during the KT tonight.

Makefile
```diff
@@ -51,12 +51,15 @@ test_helper:
...
+	@echo 'Build a docker image'
+	docker build -t provision_splunk:latest -f tests/large/provision/Dockerfile tests/large/provision
+
 	@echo 'Start container with splunk'
 	docker-compose -f tests/large/provision/docker-compose.yml up &
 
 	sleep 120
 	@echo 'Provision splunk container'
-	docker-compose -f tests/large/provision/docker-compose.yml exec -T splunk sh -c 'cd /opt/splunk;./provision.sh;/opt/splunk/bin/splunk enable listen 9997 -auth admin:changeme;/opt/splunk/bin/splunk add index test_0;/opt/splunk/bin/splunk add index test_1;/opt/splunk/bin/splunk restart'
+	docker exec --user splunk provision_splunk_1 sh -c 'cd /opt/splunk;./provision.sh;./add_httpevent_collector.sh;/opt/splunk/bin/splunk enable listen 9997 -auth admin:changeme;/opt/splunk/bin/splunk add index test_0;/opt/splunk/bin/splunk add index test_1;/opt/splunk/bin/splunk restart'
```
li-wu: You can use `--no-cache` [option](https://docs.docker.com/compose/reference/build/) here for `docker-comopse` instead of introducing more steps.

jmeixensperger: That only works for the build command, so we'd still have to build the image either way (I assume this is the extra step you are referring to).

splunk_eventgen/lib/eventgenconfig.py
```diff
@@ -520,15 +521,22 @@ def parse(self):
             if os.path.exists(s.sampleDir):
                 sampleFiles = os.listdir(s.sampleDir)
                 for sample in sampleFiles:
-                    results = re.match(s.name, sample)
+                    sample_name = s.name
+                    # If we expect a .csv, append it to the file name - regex matching must include the extension
+                    if s.sampletype == "csv" and not s.name.endswith(".csv"):
+                        sample_name = s.name + ".csv"
```
GordonWang, line 527: Maybe escape `.` will be more accurate to find the CSV sample file. 
For example, we do not want to match `abcdcsv`, however, we want to extract the `.csv` literally.

```
sample_name = s.name + "\.csv"
```

@jmeixensperger  what's your thought?

jmeixensperger: I agree, I'll test it out and make sure it works.

tests/large/provision/docker-compose.yml
```diff
@@ -2,7 +2,7 @@ version: "3.3"
 services:
   splunk:
     hostname: eventgensplunk
-    build: .
+    image: provision_splunk:latest
```
li-wu, line 5: Since docker-compose is not used, I suppose `docker-compose.yml` is not used any more?

Makefile
```diff
@@ -51,12 +51,15 @@ test_helper:
 	@echo 'Installing docker-compose'
 	bash install_docker_compose.sh
 
+	@echo 'Build a docker image'
+	docker build -t provision_splunk:latest -f tests/large/provision/Dockerfile tests/large/provision
+
 	@echo 'Start container with splunk'
 	docker-compose -f tests/large/provision/docker-compose.yml up &
```
jmeixensperger, line 58: @li-wu the docker-compose is still used here

PR #368, opened 2020-03-31
Improvement/release script

GordonWang:
send PR to develop and master at the same time


jmeixensperger:
I fixed some of the test failures that you are seeing in my PR: #365

PR #382, opened 2020-04-24
use black to format python code

li-wu:
1. Introduce [black](https://github.com/psf/black) as code format tool;
* Format all python code using `black`
```make format-all```
* Format changed or newly added python codes using `black` before commit:
```make format```

And you can still use `flake8` to lint the code:
* Lint all python code using `flake8`
```make lint-all```
* Format changed or newly added python codes using `black` before commit:
```make lint```

2. All changed python files are formatted using `black`

3. Fix some issues reported by `flake8`
* unused imports
* unused variables
* ambiguous variables

### CODE REVIEW

File: Makefile [hunk 1]
```diff
@@ -148,12 +149,12 @@ endif
...
+	@black $(NEWLY_ADDED_PY_FILES)
 endif
 
 lint-all:
 	@flake8 .
 
 format-all:
 	@isort -rc .
-	@yapf -r -i .
+	@black .
```
li-wu, line 160: Use black instead of yapf here.

pyproject.toml
```diff
@@ -0,0 +1,12 @@
+[tool.black]
```
li-wu, line 1: This is the configuration file for black.

Issue #386, opened 2020-04-29
[BUG] eventgen does not work with python 3.8


kenshinxu1:
**Describe the bug**
Follow the eventgen install document and try to install it with a new created python 3.8 venv.

a lot of errors comes up with command as below, after the switch to a newly created python3.7.5 venv, it works fine:

pip install splunk_eventgen-7.1.0.tar.gz
....
Installing collected packages: lxml, pytest-mock, docutils, jmespath, python-dateutil, botocore, s3transfer, boto3, ujson, pyyaml, httplib2, MarkupSafe, jinja2, pyflakes, pycodestyle, entrypoints, mccabe, flake8, yapf, isort, click, Werkzeug, itsdangerous, Flask, redis, requests-futures, uuid, splunk-eventgen
  Running setup.py install for lxml ... error
    ERROR: Command errored out with exit status 1:
     command: /root/eventgen/dist/eventgen-venv/bin/python3.8 -u -c 'import sys, setuptools, tokenize; sys.argv[0] = '"'"'/tmp/pip-install-m3thgm4h/lxml/setup.py'"'"'; __file__='"'"'/tmp/pip-install-m3thgm4h/lxml/setup.py'"'"';f=getattr(tokenize, '"'"'open'"'"', open)(__file__);code=f.read().replace('"'"'\r\n'"'"', '"'"'\n'"'"');f.close();exec(compile(code, __file__, '"'"'exec'"'"'))' install --record /tmp/pip-record-nppa3bp8/install-record.txt --single-version-externally-managed --compile --install-headers /root/eventgen/dist/eventgen-venv/include/site/python3.8/lxml
         cwd: /tmp/pip-install-m3thgm4h/lxml/
    Complete output (2506 lines):
    Building lxml version 4.3.4.
    Building without Cython.
    Using build configuration of libxslt 1.1.28
    Building against libxml2/libxslt in the following directory: /usr/lib64
    running install
    running build
    running build_py
    creating build
    creating build/lib.linux-x86_64-3.8
    creating build/lib.linux-x86_64-3.8/lxml
    copying src/lxml/sax.py -> build/lib.linux-x86_64-3.8/lxml
    copying src/lxml/_elementpath.py -> build/lib.linux-x86_64-3.8/lxml
    copying src/lxml/builder.py -> build/lib.linux-x86_64-3.8/lxml
    copying src/lxml/ElementInclude.py -> build/lib.linux-x86_64-3.8/lxml
[...]

Issue #381, opened 2020-04-21
[IMPROVEMENT] Add Version of httplib2 to requirements.txt


HKramm:
**Is your feature request related to a problem? Please describe.**
On my system was httplib2 in version 0.9.2 installed. It seems that in this version disable_ssl_certificate_validation=True
does not work. In my case splunkstream did not work with a self-signed certificate.

**Describe the solution you'd like**
Please add the version in the requirements.txt for httplib2. In my case the version 0.17.2 is working correctly.

jmeixensperger:
Enhancement will be included in next release.

PR #388, opened 2020-05-05
Update requirements

jmeixensperger:
Addresses #386 and #381 
lxml version is needed for python 3.8 support

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -6,7 +6,7 @@ version: 2
 jobs:
   test_eventgen:
     machine:
-      image: circleci/classic:latest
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -6,9 +6,13 @@ version: 2
 jobs:
   test_eventgen:
     machine:
-      image: circleci/classic:latest
+      image: circleci/classic:201808-01
```
jmeixensperger, line 9: 'latest' refers to an image from 2017 w/ python 3.5 installed. Updating to python 3.7.0 since we explicitly support 3.7+

Issue #387, opened 2020-04-30
[BUG] pip install fails 


roryab:
**Describe the bug**
running pip install on python 3.7 fails

Clean miniconda 3.7 venv

**To Reproduce**
/opt/miniconda3/bin/conda create -n py3 python=3.7 -y
source /opt/miniconda3/bin/activate py3
pip install git+https://www.github.com/splunk/eventgen.git

**Expected behavior**
Successful install

**Actual behavior**
```
pip install git+https://www.github.com/splunk/eventgen.git
Collecting git+https://www.github.com/splunk/eventgen.git
  Cloning https://www.github.com/splunk/eventgen.git to /tmp/pip-req-build-rppyeyvk
  Running command git clone -q https://www.github.com/splunk/eventgen.git /tmp/pip-req-build-rppyeyvk
ERROR: Exception:
Traceback (most recent call last):
  File "/opt/miniconda3/envs/py3/lib/python3.7/site-packages/pip/_internal/cli/base_command.py", line 186, in _main
    status = self.run(options, args)
  File "/opt/miniconda3/envs/py3/lib/python3.7/site-packages/pip/_internal/commands/install.py", line 331, in run
    resolver.resolve(requirement_set)
  File "/opt/miniconda3/envs/py3/lib/python3.7/site-packages/pip/_internal/legacy_resolve.py", line 177, in resolve
    discovered_reqs.extend(self._resolve_one(requirement_set, req))
  File "/opt/miniconda3/envs/py3/lib/python3.7/site-packages/pip/_internal/legacy_resolve.py", line 333, in _resolve_one
    abstract_dist = self._get_abstract_dist_for(req_to_install)
  File "/opt/miniconda3/envs/py3/lib/python3.7/site-packages/pip/_internal/legacy_resolve.py", line 282, in _get_abstract_dist_for
    abstract_dist = self.preparer.prepare_linked_requirement(req)
  File "/opt/miniconda3/envs/py3/lib/python3.7/site-packages/pip/_internal/operations/prepare.py", line 516, in prepare_linked_requirement
    req, self.req_tracker, self.finder, self.build_isolation,
  File "/opt/miniconda3/envs/py3/lib/python3.7/site-packages/pip/_internal/operations/prepare.py", line 95, in _get_prepared_distribution
    abstract_dist.prepare_distribution_metadata(finder, build_isolation)
[...]

PR #389, opened 2020-05-06
fix ujson version

li-wu:


jmeixensperger:
You might also need to update the ujson package that's being downloaded in the Makefile: https://github.com/splunk/eventgen/blob/develop/Makefile#L45

Makefile
```diff
@@ -42,7 +42,7 @@ test_helper:
 	docker exec -i ${EVENTGEN_TEST_IMAGE} /bin/sh -c "cd $(shell pwd); pip3 install dist/splunk_eventgen*.tar.gz"
 
 	@echo 'Installing test requirements'
-	docker exec -i ${EVENTGEN_TEST_IMAGE} /bin/sh -c "pip3 install --upgrade pip;pip3 install -r $(shell pwd)/requirements.txt;pip3 install git+https://github.com/esnme/ultrajson.git"
+	docker exec -i ${EVENTGEN_TEST_IMAGE} /bin/sh -c "pip3 install --upgrade pip;pip3 install -r $(shell pwd)/requirements.txt"
```
li-wu, line 45: Check related bug here: https://github.com/ultrajson/ultrajson/issues/326. This issue has been fixed in version 2.0+. So we do not need to use `git+https://github.com/esnme/ultrajson.git` to install `ujson`.

dockerfiles/Dockerfile
```diff
@@ -29,8 +29,8 @@ RUN apk --no-cache upgrade && \
 	chmod 0700 /root/.ssh && \
 	passwd -u root && \
 	# install dependencies of conduct2 used by perf
-	pip2 install filelock twisted requests queuelib ujson psutil crochet msgpack-python unidecode attrdict service_identity && \
-	pip2 install git+https://github.com/esnme/ultrajson.git
+	pip2 install filelock twisted requests queuelib psutil crochet msgpack-python unidecode attrdict service_identity && \
+	pip2 install git+https://github.com/esnme/ultrajson.git@v1.35
```
li-wu, line 33: The `conductor2` depends on `ujson < 2.0`, so we need to install it in this way.

requirements.txt
```diff
@@ -10,12 +10,11 @@ pytest-mock>=1.10.4
 boto3
 requests>=2.18.4
 requests[security]
-ujson>=1.35
+ujson==2.0.3
```
li-wu, line 13: upgrade to the latest version 2.0.3.

pyproject.toml
```diff
@@ -9,4 +9,4 @@ exclude = '''
   | build
   | dist
 )/
-```
+'''
```
li-wu, line 12: Fix typo and bug: https://github.com/splunk/eventgen/issues/387

PR #391, opened 2020-05-07
enable black py format check in circle ci

GordonWang:


.circleci/config.yml
```diff
@@ -13,6 +13,18 @@ jobs:
           name: Enable Py3
           command: |
             pyenv global 3.7.0
+      - run:
+          name: Run Flake8 Code Lint
```
GordonWang, line 17: since we have enabled py37 globally, we can move the lint and format check process out of container.

Makefile
```diff
@@ -12,6 +12,7 @@ LARGE ?= 'tests/large'
 XLARGE ?= 'tests/xlarge'
 NEWLY_ADDED_PY_FILES = $(shell git ls-files -o --exclude-standard | grep -E '\.py$$')
 CHANGED_ADDED_PY_FILES = $(shell git ls-files -mo --exclude-standard | grep -E '\.py$$')
+ALL_PY_FILES = $(shell git ls-files tests splunk_eventgen | grep -E '\.py$$')
```
GordonWang, line 15: only lint and format py files in tests and splunk_eventgen folder.

GordonWang:
This PR has a bunch of py string changes. In order to make sure all the files can pass flake8 line length lint rules, I made these changes. 
There should be no impact on functionality. Only code format changes.

Issue #394, opened 2020-05-15
[FEATURE/IMPROVEMENT] Set default source based on sample file name


jmeixensperger:
**Is your feature request related to a problem? Please describe.**
I want my event 'source' to contain the sample file name, but hard-coding each file name / maintaining a conf stanza for each file is inefficient.

**Describe the solution you'd like**
I want eventgen to use the sample file name by default.

**Describe alternatives you've considered**
Passing a special string / additional option to enable this feature, although the current default 'eventgen' source can still be used by adding the option in your eventgen.conf.

**Additional context**
cc @rfaircloth-splunk feel free to elaborate

PR #395, opened 2020-05-15
default source

jmeixensperger:
For #394 
Changes the default source behavior to use the sample file name. Specifying a source in eventgen.conf still has the same behavior, and will prevent this.

### DIFF
--- a/splunk_eventgen/lib/eventgenconfig.py
+++ b/splunk_eventgen/lib/eventgenconfig.py
@@ -971,7 +971,7 @@ class Config(object):
                 s.generator = "replay"
             # 5/14/20 - Instead of using a static default source, leave source empty by default and
             # set it to the sample file name unless otherwise specified.
-            if not self.source:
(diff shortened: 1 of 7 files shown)


splunk_eventgen/lib/eventgenconfig.py
```diff
@@ -971,6 +969,11 @@ def parse(self):
                 s.interval = 0 if not s.interval else s.interval
                 # 12/29/13 CS Moved replay generation to a new replay generator plugin
                 s.generator = "replay"
+            # 5/14/20 - Instead of using a static default source, leave source empty by default and
+            # set it to the sample file name unless otherwise specified.
+            if not self.source:
+                sample_path = s.filePath if s.filePath else s.generator
```
jmeixensperger, line 975: The only scenario I found in which s.filePath is not populated is when a generator (i.e. windbag) was used instead of a sample file.

PR #409, opened 2020-07-29
remove config copy for perf regression

jmeixensperger:
for perf regression in 7.0.0 -> 7.1.0
I built a fresh package from the 7.1.0 branch with the exception of these changes and reached my 400GB/day target rate that 7.1.0 could not achieve.


jmeixensperger:
I removed the uuid dependency (which was blocking our CI from installing the .tar.gz package) and reverted the fixed ujson version that the performance team does not need anymore (confirmed w/ Yifeng).

PR #413, opened 2020-10-21
perdayvolume fix

jmeixensperger:
- abstracted token replacement out of generatorplugin.py for plugin-specific handling
- EDIT: preview difference in token vs. replaced size, update target interval size with ratio
- fix 2x events from duplicate queue_it
- resolve kramdown dependency alert

### DIFF
--- a/docs/Gemfile.lock
+++ b/docs/Gemfile.lock
@@ -67,7 +67,7 @@ GEM
       jekyll-theme-time-machine (= 0.1.1)
       jekyll-titles-from-headings (= 0.5.1)
       jemoji (= 0.10.2)
-      kramdown (= 1.17.0)
+      kramdown (>= 2.3.0)
       liquid (= 4.0.0)
       listen (= 3.1.5)
       mercenary (~> 0.3)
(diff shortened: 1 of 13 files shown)


lephino (approved):
We should look at replacing that the event object is all in memory.  If someone does a large perday volume on an interval, we will have issues.  We should probably change perdayvolume to use an in memory buffered io stream.

jmeixensperger:
> We should look at replacing that the event object is all in memory. If someone does a large perday volume on an interval, we will have issues. We should probably change perdayvolume to use an in memory buffered io stream.

Which 'event object' are you referring to? The eventsDict that we compile? The send_objects that we see after replacing tokens? Not super experienced with these kind of optimizations, so trying to understand fully.

lephino (approved):

lephino:
> > We should look at replacing that the event object is all in memory. If someone does a large perday volume on an interval, we will have issues. We should probably change perdayvolume to use an in memory buffered io stream.
> 
> Which 'event object' are you referring to? The eventsDict that we compile? The send_objects that we see after replacing tokens? Not super experienced with these kind of optimizations, so trying to understand fully.

Specifically the dictionary that has all the events - def send_events(self, send_objects, startTime): - the send_objects is currently built in memory with just a dictionary.  Long term we should think about moving this to an in-memory file player.  That will allow us to cache and move forward / backward with large items without creating a "mem leak".  Or we can just tell our end users you need more ram ;P

PR #433, opened 2021-07-02
Created pypy_eventgen image + enabled MacOS builds

KarolOlko:


KarolOlko:
Did some throughput testing and this pypy-powered version of eventgen was not only consuming much more memory, but generating less load! This is a surprising result, but I'm glad I tried. Nevertheless it doesn't make sense to integrate this, so I'm closing the PR.

PR #441, opened 2022-05-09
Update replay generator

### LINKED ISSUE #440: [BUG] Replay mode timeline bug
**Describe the bug**
Replay mode appears to have a bug whereby the first two events are replayed at the same time, and for each event thereafter the time offset applied from one event to the next is off by 1 event.

**To Reproduce**
1. Create a basic replay config (see below)
2. Create a simple list of events (see example below)
3. Note that the first two events are replayed at the same time, and every subsequent events offset is incorrect.

**Expected behavior**
Each log should be replayed 5 seconds apart, except the 5th log entry should be replayed 5 minutes after the previous
```
Jan 17 00:00:00 TEST 1
Jan 17 00:00:05 TEST 2
Jan 17 00:00:10 TEST 3
Jan 17 00:00:15 TEST 4
Jan 17 00:05:15 TEST 5
```

**Actual behavior**
The first two logs are replayed with the same time, all subsequent logs are offset by the previous logs offset.
```
Jan 17 00:00:00 TEST 1
Jan 17 00:00:00 TEST 2
Jan 17 00:00:05 TEST 3
Jan 17 00:00:10 TEST 4
Jan 17 00:00:15 TEST 5
Jan 17 00:05:15 TEST 6
```

**Sample files and eventgen.conf file**
eventgen.conf
```
[test.log]
mode = replay
outputMode = stdout

token.0.token = \w{3}\s+\d{1,2}\s+\d{2}:\d{2}:\d{2}
token.0.replacementType = replaytimestamp
token.0.replacement = %b %d %H:%M:%S
```
test.log
```
Jan 17 00:00:00 TEST 1
Jan 17 00:00:05 TEST 2
Jan 17 00:00:10 TEST 3
Jan 17 00:00:15 TEST 4
Jan 17 00:05:15 TEST 5
```

**Do you run eventgen with SA-eventgen?**
No

**If you are using eventgen with pip module mode (please complete the following information):**
 - python version: 3.8
 - OS: Ubuntu 20.04 & Alpine 3.13
 - Virtual Env is used: Yes
 - Eventgen Version 7.2.1

breakfixrepeat:
As promised, resolves #440

### DIFF
--- a/splunk_eventgen/lib/plugins/generator/replay.py
+++ b/splunk_eventgen/lib/plugins/generator/replay.py
@@ -173,7 +173,10 @@ class ReplayGenerator(GeneratorPlugin):
                     rpevent, self.backfill_time, earliest, latest
                 )
                 previous_event = current_event
-                previous_event_timediff = line_list[index + 1]["timediff"]
+                if len(line_list) != 1:
+                    previous_event_timediff = line_list[index + 1]["timediff"]
+                else:
+                    previous_event_timediff = rpevent["timediff"]
                 self._out.bulksend([current_event])
                 continue
             else:
--- a/splunk_eventgen/lib/plugins/generator/replay.py
+++ b/splunk_eventgen/lib/plugins/generator/replay.py
@@ -173,9 +173,12 @@ class ReplayGenerator(GeneratorPlugin):
                     rpevent, self.backfill_time, earliest, latest
                 )
                 previous_event = current_event
-                previous_event_timediff = rpevent["timediff"]
+                previous_event_timediff = line_list[index + 1]["timediff"]
                 self._out.bulksend([current_event])
                 continue
+            else:
+                previous_event = rpevent
+                previous_event_timediff = rpevent["timediff"]
             try:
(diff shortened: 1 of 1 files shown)


splunk_eventgen/lib/plugins/generator/replay.py
```diff
@@ -173,9 +173,12 @@ def gen(self, count, earliest, latest, samplename=None):
                     rpevent, self.backfill_time, earliest, latest
                 )
                 previous_event = current_event
-                previous_event_timediff = rpevent["timediff"]
+                previous_event_timediff = line_list[index + 1]["timediff"]
```
0xr2po: If you have a jsonl file with one line, you would run into a problem with that. I suggest something like:
```suggestion
                if len(line_list) != 1:
                    previous_event_timediff = line_list[index + 1]["timediff"]
                else:
                    previous_event_timediff = rpevent["timediff"]
```

breakfixrepeat: Nice catch! Fixed

0xr2po (approved):
Approved

